PR #4, opened 2022-04-14
jcfg:make umem regions dynamic size

KeithWiles:
The UMEM region size was fixed at 16, we need more than 16.
Convert the jcfg_umem_t.rinfo structure array to an allocated array
based on region count.
Increased max number of regions to 128.

### DIFF
--- a/lib/usr/app/jcfg/jcfg.h
+++ b/lib/usr/app/jcfg/jcfg.h
@@ -29,7 +29,7 @@ extern "C" {
 #endif
 
 #define DEFAULT_CHUNK_SIZE   1024
-#define UMEM_MAX_REGIONS     16
+#define UMEM_MAX_REGIONS     128
 #define JCFG_MAX_STRING_SIZE 32
 
 #include <cne_common.h>        // for CNDP_API, CNE_STD_C11
@@ -154,9 +154,8 @@ typedef struct jcfg_umem {
     uint16_t txdesc;            /**< Number of Tx descriptors */
     uint16_t idx;               /**< The UMEM index id 0 to N */
     uint16_t shared_umem;       /**< Enable shared umem support */
-
-    uint16_t region_cnt;                   /**< Number of regions defined */
-    region_info_t rinfo[UMEM_MAX_REGIONS]; /**< Region information data */
+    uint16_t region_cnt;        /**< Number of regions defined */
+    region_info_t *rinfo;       /**< Region information data */
 } jcfg_umem_t;
 
 /**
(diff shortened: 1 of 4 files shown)


lib/usr/app/jcfg/jcfg_private.h
```diff
@@ -90,6 +90,15 @@ jcfg_get_json_token(struct jcfg *cfg)
...
 }
 
+/**
+ * Free the internal umem structure.
+ * @internal
+ *
+ * @param hdr
+ *   The generic header structure pointer.
+ */
+void jcfg__umem_free(jcfg_hdr_t *hdr);
```
jeffreybshaw, line 100: Why use the double underscore here? Is it becuase its private? I don't think its necessary, and is not the convention used anywhere else.

lib/usr/app/jcfg/jcfg_umem.c
```diff
@@ -181,3 +187,12 @@ _decode_umems(struct json_object *obj, int flags, struct json_object *parent __c
 
     return ret ? ret : JSON_C_VISIT_RETURN_SKIP;
 }
+
+void
+jcfg__umem_free(jcfg_hdr_t *hdr)
+{
+    jcfg_umem_t *umem = (jcfg_umem_t *)hdr;
+
+    free(umem->rinfo);
```
jeffreybshaw, line 196: Can you check if umem is null here?

lib/usr/app/jcfg/jcfg_umem.c
```diff
@@ -181,3 +187,12 @@ _decode_umems(struct json_object *obj, int flags, struct json_object *parent __c
...
+    mmap_free(umem->mm);
+}
```
jeffreybshaw, line 198: can you set

```
umem->rinfo = NULL;
umem->mm = NULL;
```

lib/usr/app/jcfg/jcfg.h
```diff
@@ -154,9 +154,9 @@ typedef struct jcfg_umem {
     uint16_t txdesc;            /**< Number of Tx descriptors */
     uint16_t idx;               /**< The UMEM index id 0 to N */
     uint16_t shared_umem;       /**< Enable shared umem support */
-
-    uint16_t region_cnt;                   /**< Number of regions defined */
-    region_info_t rinfo[UMEM_MAX_REGIONS]; /**< Region information data */
+    uint16_t region_cnt;        /**< Number of regions defined */
+    uint16_t region_size;       /**< Size of the rinfo array for realloc() calls */
```
jeffreybshaw, line 158: where is region_size used?

jeffreybshaw (changes requested):

lib/usr/app/jcfg/jcfg_umem.c
KeithWiles: umem can not be NULL here as it is only called from one spot.

KeithWiles:
Need to squash the commits before merge.

jeffreybshaw (approved):

PR #11, opened 2022-04-19
jsonc file formatting

elzamath:
When using the tools/check_json.py script to check the various jsonc
files in the repo, came across an error of the form:

file cannot be parsed (message: Expecting property name enclosed in double quotes..)

These changes get rid of the error.

### DIFF
--- a/examples/cnet-quic/quic.jsonc
+++ b/examples/cnet-quic/quic.jsonc
@@ -126,7 +126,7 @@
         "cli": true,
 
         // Optional example application options, note the key-name needs to match the thread name
-        "graph:0": ["ip4*", "udp*", "tcp*", "pkt_drop", "chnl*", "kernel_recv*"],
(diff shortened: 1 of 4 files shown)


examples/cnet-quic/quic.jsonc
```diff
@@ -126,7 +126,7 @@
         "cli": true,
 
         // Optional example application options, note the key-name needs to match the thread name
-        "graph:0": ["ip4*", "udp*", "tcp*", "pkt_drop", "chnl*", "kernel_recv*"],
+        "graph:0": ["ip4*", "udp*", "tcp*", "pkt_drop", "chnl*", "kernel_recv*"]
```
maryamtahhan, line 129: If there's no significance to the number other than just an id in the graph naming `graph:0` I would be inclined to comment out this line and leave the `graph:1` active as if I ever want to enable `graph:0` then I just need to uncomment a line and not remember to add the comma at the end.

or if `graph:1` is not needed then maybe remove it

maryamtahhan (approved):

KeithWiles (approved):

PR #16, opened 2022-04-26
cndpgo ipv4 checksum verify example, module name changes

dakshinai:
Signed-off-by: Dakshina Ilangovan <dakshina.ilangovan@intel.com>

Two changes 
1. Added a flow to receive packets, compute and verify checksum in packet if ipv4 and drop
The summation loop reduces performance
2. Changed cndpgo module names

### DIFF
--- a/lang/go/cndp.org/cndpgo/const.go
+++ b/lang/go/cndp.org/cndpgo/const.go
@@ -17,6 +17,7 @@ const (
 // is not determined inside packet. Only default minimum size is used.
 const (
 	EtherLen = 14
+	IPv4Len  = 20
 	IPv6Len  = 40
 )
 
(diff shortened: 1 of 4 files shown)


dakshinai:
Also, 

1. We need to fix on go version. The code requires 1.16 and I have tested it with 1.17. Please let me know what version the other packages require?
2. Need to rename module names and directory name for other golang packages, do we need all of them, not sure which are being used for Prometheus e.t.c
3. cndpgo module can be published for go tools. It can have its own version number.

KeithWiles (commented):
Why did we need to add a checksum mode, did a customer request this to be done in Go?

dakshinai:
> Why did we need to add a checksum mode, did a customer request this to be done in Go?

This was Edwin's request to mimic a customer's flow. We will be reporting performance numbers for this flow as well.

Please let me know if this can be in standard examples or I will have to use them anyway to get some numbers.

PR #17, opened 2022-04-27
Rust bindings and high level library for CNDP CNE.

manojgop:
- Add cne-sys rust API bindings for CNP CNE C library.
 - Add cne high level rust API library based on cne-sys.
 - Add test cases in rust cne-sys and cne library crate.
 - Example loop back application which uses cne rust library APIs.
 - Sample scripts to run test cases and example application.
 - Add README.md file.


lang/rs/cndprs/fwd.json
```diff
@@ -0,0 +1,64 @@
+{
```
KeithWiles, line 1: Does Rust not have a json-c package?

lang/rs/cndprs/src/bindings/bindings.h
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+#include <stdint.h>         // for uint64_t, uint32_t, uint8_t
+#include <pktmbuf.h>        // for pktmbuf_pool_create, pktmbuf_info_t
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+uint16_t _pktdev_rx_burst(uint16_t lport_id, pktmbuf_t **rx_pkts, const uint16_t nb_pkts);
```
KeithWiles, line 21: Not a fan of _pktdev_rx_burst() having the _ in front of the names seems misleading to me.  Maybe pktdev_rx_burst_fn() or something like it.

lang/rs/cndprs/src/bindings/bindings.h
```diff
@@ -0,0 +1,35 @@
...
+
+uint16_t _pktdev_tx_burst(uint16_t lport_id, pktmbuf_t **tx_pkts, uint16_t nb_pkts);
```
KeithWiles, line 23: Do these need the CNDP_API macro with the prototypes?

KeithWiles (changes requested):
This looks good, but need to have a call to better understand the code. I we need to adjust the naming and directory layout to better align with future needs. I did not review the code per say, so that is why we need to the conf call.

lang/rs/cndprs/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2020-2022 Intel Corporation.
+
+[package]
+name = "cndprs"
+version = "0.1.0"
+authors = ["Manoj Gopalakrishnan <manoj.gopalakrishnan@intel.com>"]
```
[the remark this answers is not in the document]
manojgop: I remove the authors.

lang/rs/cndprs/build.rs
```diff
@@ -0,0 +1,77 @@
...
+ */
+extern crate bindgen;
+
+use std::env;
+use std::fs;
+use std::path::PathBuf;
+
+fn main() {
+    // PKG_CONFIG_PATH environment variable should be set to directory containing libcndp.pc file.
+    let cndp_lib_dir = pkg_config::get_variable("libcndp", "libdir").unwrap();
```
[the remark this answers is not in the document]
manojgop: This assumes libcndp.pc file is present in PKG_CONFIG_PATH  and extracts CNDP library and include file path from libcndp.pc file. This is a standard way of using pkgconfig. But if we want to use some default system path for CNDP library and include path we can do that as well

lang/rs/cndprs/src/bindings/bindings.h
KeithWiles: OK, we can leave it this way.

lang/rs/bindings/cners/examples/loopback/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+# SPDX-License-Identifier: BSD-3-Clause
```
KeithWiles, line 1: Can this move to lang/rs/bindings/examples ??

lang/rs/bindings/cners/examples/loopback/run.sh
```diff
@@ -0,0 +1,21 @@
...
+
+# Build CNDP and install (sudo CNE_DEST_DIR=/ make install) before running this script.
+# Usage: ./run.sh <json config file> <port name> <core id (optional)>
+# eg: ./run.sh ../../fwd.json enp134s0:0 25
+
+# JSON file
+CONFIG=${1:-"../../fwd.json"}
+
+# Lport
+PORT=${2:-"enp134s0:0"}
```
KeithWiles, line 11: We already have the port name in the json file, lets not duplicate it here too. Cores are also defined in the json file.

lang/rs/bindings/cners/examples/loopback/src/main.rs
```diff
@@ -0,0 +1,221 @@
...
+    match cne.configure(&jsonc_file) {
+        Err(e) => {
+            log::error!("Error configuring CNE: {}", e.to_string());
+            return;
+        }
+        Ok(_) => {}
+    }
+
+    // Get CNE port to recv/send packets.
+    let port = cne.get_port(&port_name);
```
KeithWiles, line 94: We should not be referencing ports by name per say and all of the CNDP APIs use port number. I assume the port here is just a number from 0 - N.

The ports names in the application should not be used other than printing out information. This way we do not need to update the command line or hardcode the names in the application.

lang/rs/bindings/cners/examples/loopback/src/main.rs
```diff
@@ -0,0 +1,221 @@
...
+
+        CneInstance::unregister_thread(tid).unwrap();
+    });
+
+    // log port stats. Wait for ctrl-c event from user.
+    loop {
+        select! {
+            recv(ticks) -> _ => {
+                if running.load(Ordering::SeqCst) {
+                    let port = cne.get_port(&port_name).unwrap();
```
KeithWiles, line 146: Port name again, just use port number. Plus this is a lookup I am guessing and a lookup may consume cycles.

We also need to support multiple ports in this application, but that can be done after these are address in a different PR.

lang/rs/bindings/cners/src/config.rs
```diff
@@ -0,0 +1,518 @@
...
+
+    pub fn cleanup(&mut self) -> Result<(), CneError> {
+        self.cleanup_lports()?;
+
+        self.cleanup_mem_pool()?;
+
+        Ok(())
+    }
+
+    pub fn get_port_by_name(&self, port_name: &str) -> Result<Port, CneError> {
```
KeithWiles, line 178: Not sure we need get_port_by_name() if we are using port number. something like get_port_info(int port_id) would be more useful IMO.

lang/rs/bindings/cners/src/config.rs
```diff
@@ -0,0 +1,518 @@
...
+        ret
+    }
+
+    fn setup_mem_pool(&mut self) -> Result<(), CneError> {
+        for (umem_name, umem) in &mut self.umems {
+            umem.name = umem_name.clone();
+            umem.bufcnt *= 1024;
+            umem.bufsz *= 1024;
+
+            if umem.rxdesc == 0 {
```
KeithWiles, line 207: Does rust require {} around single line if, while, for, ... statements?

Does Rust have a formatter? If the formatter add them it is OK.

lang/rs/bindings/cners/src/config.rs
```diff
@@ -0,0 +1,518 @@
...
+            }
+            umem.rxdesc *= 1024;
+
+            if umem.txdesc == 0 {
+                umem.txdesc = self.defaults.as_ref().map(|d| d.txdesc).unwrap_or(0);
+            }
+            umem.txdesc *= 1024;
+
+            unsafe {
+                let default_mmap_type = String::from("4KB");
```
KeithWiles, line 218: Are we forcing 4KB pages here? If so then we need to use the json file request.

lang/rs/bindings/cners/src/error.rs
```diff
@@ -0,0 +1,32 @@
...
+        let s = match self {
+            CneError::BufferAllocError(s) => s,
+            CneError::ConfigError(s) => s,
+            CneError::PortError(s) => s,
+            CneError::PortStatsError(s) => s,
+            CneError::RegisterError(s) => s,
+            CneError::RxBurstError(s) => s,
+            CneError::TxBurstError(s) => s,
+        };
+        write!(f, "CndpError({})", s)
```
KeithWiles, line 30: CndpError should be CneError ??

lang/rs/bindings/cners/src/instance.rs
```diff
@@ -0,0 +1,134 @@
...
+
+        Ok(())
+    }
+
+    pub fn cleanup(&self) -> Result<(), CneError> {
+        let mut cndp = self.write();
+        if let Some(cfg) = &mut cndp.cfg {
+            cfg.cleanup()?;
+        }
+        cndp.cfg = None;
```
KeithWiles, line 78: cndp reference

lang/rs/bindings/cners/src/lib.rs
```diff
@@ -0,0 +1,323 @@
...
+
+        // Get CNE instance.
+        let cne = CneInstance::get_instance();
+
+        // Configure CNE.
+        let ret = cne.configure(&jsonc_file);
+        assert!(ret.is_ok());
+
+        // Get valid port. Should return success.
+        let port = cne.get_port("enp134s0:0");
```
KeithWiles, line 128: hard coded port name again.

lang/rs/bindings/cners/src/lib.rs
```diff
@@ -0,0 +1,323 @@
...
```
KeithWiles, line 241: port name

lang/rs/bindings/cners/src/port.rs
```diff
@@ -0,0 +1,107 @@
...
+ */
+
+use super::bindings::bindings::{
+    _pktdev_rx_burst, _pktdev_tx_burst, lport_stats_t, pktdev_buf_alloc, pktdev_stats_get,
+};
+use super::error::*;
+use super::packet::*;
+use std::mem::MaybeUninit;
+
+pub struct Port {
```
KeithWiles, line 12: This is similar to the Go code and we do not need a structure for just a port number. This needs to be fixed in the Go code too.

KeithWiles (commented):
We needs some cleanup for cndp naming and remove the need to know the port name and instead use port number.

We can address some of these in a new PR if you like.

lang/rs/bindings/cners/examples/loopback/Cargo.toml
manojgop: There are examples where Rust bindings for C/C++ library ends in -sys or rs. We could rename it as well.

lang/rs/bindings/cners/examples/loopback/src/main.rs
manojgop: same as https://github.com/CloudNativeDataPlane/cndp/pull/17#discussion_r859992279

lang/rs/bindings/cners/examples/loopback/src/main.rs
manojgop: Multiple ports in applicationc an be different PR. I want a simple example as part of library crate. We can have example applications outside library crate

lang/rs/bindings/cners/src/config.rs
manojgop: It's an option. We can have both IMO.  I feel using port name in JSON file is more user friendly than counting the port index in lport section in JSON.

lang/rs/bindings/cners/examples/loopback/run.sh
jeffreybshaw: All the ports in the json file should be used for send/recv, right? I agree with Keith, why would somebody put something in the json file they do not want to use?

lang/rs/bindings/cners/src/instance.rs
```diff
@@ -0,0 +1,134 @@
...
+                CNDPINSTANCE.write(singleton);
+            });
+
+            // Return shared reference to the cndp instance, safe for concurrent access.
+            CNDPINSTANCE.assume_init_ref()
+        }
+    }
+}
+
+// CndpInstance public functions.
```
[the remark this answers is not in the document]
manojgop: Will change

lang/rs/bindings/cners/src/lib.rs
manojgop: This is a unit test code in lib.rs.

lang/rs/bindings/cners/src/port.rs
manojgop: I used Port is an abstract data type used to represent an CNE lport. This has both data and functions just like a C++ class which allows to send/recv packets from an lport.

lang/rs/bindings/cners/examples/loopback/run.sh
manojgop: In JSON file, user can give all ports. But this simple example application will send/recv only from one port at a time. So in command line user has to specify either port id or port name from the list. Keith and I just talked. We agreed to replace port name with port id since port name can change across systems.

lang/rs/bindings/cners/examples/loopback/Cargo.toml
manojgop: Pushed changes. User can search with `cndp_binding` in Rust crates.io (whenever we publish it. This is package name). In code we will use crate name `cne_binding`. Code is kept under `lang/rs/bindings/cne`

KeithWiles (commented):
These changes seem OK, but we need to review the code in a conference call.

lang/rs/bindings/cners/examples/loopback/run.sh
manojgop: Regarding affinitizing thread to core, Rust applications mostly uses it's own thread model and library to create and manage thread and doesn't use CNE thread APIs. So currently the "threads" section in JSON file is unused. So after Rust applications creates loopback thread, to get best performance (similar to C CNDP application) it can (optionally) set core affinity.  But we can certainly provide a Rust API to assign an lcore-group to Rust thread and there by setting core affinity to that thread based on configuration in JSON file. Just that I've not implemented it yet.

lang/rs/cndprs/fwd.json
manojgop: I tried it and it works. But Rust JSONC crate (Hjson) depends on older version (0.8) of serde JSON and won't have some newer features of serde JSON.  Also newer version of serde JSON is stable version (1.0).  If we prefer JSONC file then using Hjson which depends on older version of serde is an option. I'll update PR to use Hjson.

lang/rs/bindings/cne/examples/loopback/src/main.rs
```diff
@@ -0,0 +1,221 @@
...
+                .takes_value(true)
+                .help("CNE JSON config file"),
+        )
+        .arg(
+            Arg::with_name("port")
+                .short("p")
+                .long("port")
+                .required(true)
+                .takes_value(true)
+                .help("port name in JSON file"),
```
KeithWiles, line 38: Did this one change to a number and not a name?

KeithWiles (changes requested):

lang/rs/bindings/cne/examples/loopback/src/main.rs
manojgop: correct. I'll update it to port index

KeithWiles (approved):
LGTM, do we need a code review as you are the only one that knows Rust per say.

Will let Jeff approve or not, before merging.

manojgop:
> LGTM, do we need a code review as you are the only one that knows Rust per say.
> 
> Will let Jeff approve or not, before merging.

I would recommend code walk through. I'll schedule it early next week.

manojgop:
There is one problem with indexing by port id in JSONC file if we want to preserve the order of lports entries. JSON serializations/deserialization using HashMap doesn't maintain the order. So port name "a" and "b" enter in that order in JSONC file should index to port id 0 and 1. But since JSONC serialization using HashMap doesn't maintain order, port name and index need not be in order for eg: `a= lport 1` and `b = lport 0`. But searching lport index by lport name does not have this problem. Another option is to use a ordered map data structure like LinkedHashMap. But then serde supports this only with latest version. But then LinkedHashMap cannot be used with HJson since it is not compatible with new version of serde :-)

lang/rs/bindings/cners/examples/loopback/run.sh
KeithWiles: Also the thread section also helps the application define what a thread is doing in the application. I know this does not apply the Rust threads, but we can not forget this point. Also the thread section also assigns lport to threads or functions i.e. fwd:0, fwd:1 how do we arrange the lports to groups of threads in Rust?

manojgop: Rust applications using CNE Rust library can create Rust threads using standard routines and register those threads with CNE and we can send/recv packets using lports from that threads. The synchronization/thread safety when using same lports from multiple threads is taken care by Rust CNE library.  The loopback sample application currently doesn't use Threads section defined in JSONC file. Not sure if that answer above question. If not, I've not understood the requirement well. We can discuss in a call.

PR #22, opened 2022-04-28
go: rework directory layout

KeithWiles:
Rework the lang/go directory layout, remove some old components and
rename the binding layer to cne instead of cndpgo. The rename of the modules
to be inline with standard go applications on github.
Rerun go mod tidy making sure we have go.mod files correct for the
new layout. The go.sum files should be included in the git repo.

I have ran both bindings/examples/fwd and tool/cmon to verify they work need
to verify metrics and prometheus work in a container or k8s. The new bindings
directory will contain other binding layers i.e. hyperscan, multi-buffer IPsec, ...

```
lang/go
├── bindings
│   ├── cne        # renamed from cndpgo to cne as we should not use cndp.
│   └── examples
│       └── fwd
├── stats
│   ├── metrics
│   └── prometheus
└── tools
    ├── cmon
    └── pkgs
        ├── asciichart
        │   ├── cmd
        │   └── examples
        ├── colorize
        ├── etimers
        ├── graphdata
        ├── hexdump
        ├── irq
        ├── taborder
        ├── ttylog
        └── utils

```
One areas that was addressed was the naming of modules to align with
the public github.com directory structure. The github.com account is
call github.com/CloudNativeDataPlane, which means we need to rename
the modules. The CNDP repo on github.com will be called 'cndp' i.e.
 --> cndp.org/cmon is now github.com/CloudNativeDataPlane/cndp/lang/go/tools/cmon

Modified lang/go/stats/metrics to not require ttylog, which makes this
module not dependent on lang/go/tools/pkgs/ttylog.

For now we can leave cmom in the CNDP repo, but it may be moved to a new repo
in the future. I would like to make better integrated into CNDP as a tool to
help configure and understand the configuration of CNDP application and the system.


lang/go/.gitignore
```diff
@@ -0,0 +1,3 @@
+# Remove some binary applications
```
dakshinai, line 1: I noticed that prometheus package created a binary on go build, that can be added here as well.

dakshinai:
The current paths to modules like 
github/cloudnativedataplane/cndp/lang/go/bindings/cne is fine.

When we publish we could look at a shorter path. Versioning can be added to the published modules.

Looks good to me.

jeffreybshaw (approved):
Let's merge this.

PR #26, opened 2022-04-29
Add docker image build action

jeffreybshaw:
Signed-off-by: Jeff Shaw <jeffrey.b.shaw@intel.com>


jeffreybshaw:
This PR adds an action to verify the docker image builds. It fails because PR https://github.com/CloudNativeDataPlane/cndp/pull/25 is not merged yet.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,20 @@
+name: Docker Image Build
+
+on:
+  pull_request:
+    branches: [ main ]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
```
elzamath, line 8: should we also add this `-${{ github.event.pull_request.number }}` in case multiple PRs trigger this action at the same time

jeffreybshaw: I did some digging, and I think we're OK. This [doc](https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables) mentions for pull-requests the GITHUB_REF is set to "refs/pull/<pr_number>/merge", so we this workflow should only cancel in progress workflows for a given PR.

elzamath (approved):

PR #27, opened 2022-04-29
fix possible null pointer dereference

elzamath:
The pointer returned by the call to inet_ntop4 could be NULL
and will be dereferenced by the call to snprintf. Added a NULL
pointer check to avoid that.


lib/cnet/pcb/cnet_pcb.c
```diff
@@ -191,15 +191,15 @@ pcb_show(mempool_t *mp __cne_unused, void *obj_cb_arg __cne_unused, void *obj,
     cne_printf("[green]%-6s [orange] %04x [red]%6s[]", pcb->closed ? "Closed" : "Open",
                pcb->opt_flag, chnl_protocol_str(pcb->ip_proto));
 
-    if (snprintf(fbuf, sizeof(fbuf), "%s:%d",
-                 inet_ntop4(ip1, sizeof(ip1), &pcb->key.faddr.cin_addr, NULL),
+    b1 = inet_ntop4(ip1, sizeof(ip1), &pcb->key.faddr.cin_addr, NULL);
```
KeithWiles, line 194: Did you look at the inet_ntop4 function to do this test. Maybe that is a better location to set the invalid string text?

The inet_ntop4 routine is used in a number of places and we would need to fix all of them.

jeffreybshaw: I don't like the idea of returning a const string. Then the user doesn't know if there was an error.

This issue is found by klocwork. It only finds these two issues because the result if inet_ntop4() is passed to snprintf() where all the other usages the possible NULL pointer is passed to cne_printf(), which klocwork apparently is ok with.

lib/cnet/pcb/cnet_pcb.c
```diff
@@ -191,15 +191,15 @@ pcb_show(mempool_t *mp __cne_unused, void *obj_cb_arg __cne_unused, void *obj,
...
+    if (snprintf(fbuf, sizeof(fbuf), "%s:%d", (b1) ? b1 : "Invalid IP",
```
jeffreybshaw, line 195: You don't need the parenthesis around the (b1).

You also don't need two variables, b1 and b2. And actually you don't need both fbuf and lbuf and ip1 and ip2.
Could be simpler, IMO.

```
    char buf[128], ipb[IP4_ADDR_STRLEN] = {0}, *ip;
    uint16_t port;

    ip = inet_ntop4(ipb, sizeof(ipb), &pcb->key.faddr.cin_addr, NULL);
    port = ntohs(CIN_PORT(&pcb->key.faddr));

    if (snprintf(buf, sizeof(buf), "%s:%d", ip ?: "Invalid IP", port) < 0)
        CNE_RET("Truncated buffer data\n");
    cne_printf(" [orange]%20s[]", buf);

    ip = inet_ntop4(ipb, sizeof(ipb), &pcb->key.laddr.cin_addr, NULL);
    port = ntohs(CIN_PORT(&pcb->key.laddr));

    if (snprintf(buf, sizeof(buf), "%s:%d", ip ?: "Invalid IP", port) < 0)
        CNE_RET("Truncated buffer data\n");
    cne_printf(" [orange]%20s[]", buf);
```

lib/cnet/pcb/cnet_pcb.c
elzamath: should I fix the other places where inet_ntop4() is used even though klockwork doesn't report them? cne_printf() doesn't have a null pointer check so I am assuming we still need to fix it.. 
I was actually looking at usrtools/txgen/[app/stats|cmds|pcap].c for inet_ntop4 usage so that's why I decided to handle the NULL pointer check this way.

lib/cnet/pcb/cnet_pcb.c
elzamath: will remove the parenthesis and use one variable

lib/cnet/pcb/cnet_pcb.c
elzamath: > Did you look at the inet_ntop4 function to do this test. 

I was actually looking at https://github.com/CloudNativeDataPlane/cndp/blob/69605b8ac18723a691da1dbfce1bce74f69f86eb/usrtools/txgen/app/cmds.c#L87 to see how the NULL pointer check was handled

> This issue is found by klocwork. It only finds these two issues because the result if inet_ntop4() is passed to snprintf() where all the other usages the possible NULL pointer is passed to cne_printf(), which klocwork apparently is ok with.

Even though klockwork doesn't report the other issues, should we fix them in this PR or create a separate one for those?

jeffreybshaw: Let's fix these ones first so we get back to a cleaner klocwork report.

We should fix the other ones in another PR. We should also manually audit the rest of the source code where we might pass NULL pointers as arguments to formats expecting "%s" as this results in undefined behavior according to C.

jeffreybshaw (approved):

lib/cnet/pcb/cnet_pcb.c
elzamath: https://github.com/CloudNativeDataPlane/cndp/pull/35 - created another PR here with the fixes for the other places where inet_ntop4 is used

PR #29, opened 2022-05-02
vec:remove libvec

KeithWiles:
Remove the need for a libvec and move the cne_vec.h into lib/include directory.
Also removed the vec_pool* and vec_histogram* code as it was not used. Part of
this change is for later updates for CNET/TCP.


jeffreybshaw (approved):
Can you also remove the reference to "@TOPDIR@/lib/usr/clib/vec" from teh doc/api/doxy-api.conf.in file? This can be an additional commit or future (soon, hopefully so we don't forget) PR.

KeithWiles:
Merge with last commit in series.

PR #30, opened 2022-05-02
vec:cleanup vec header includes

KeithWiles:
Remove the cne_vec.h includes that were not required in some files.


jeffreybshaw (approved):
Thank you for organizing the PR this way. Very easy to review just this commit, but the commit message has a minor typo.

```
Remove the cne_vec.h includes at were not required in some files.
```

I think teh "at" should be "that". Minor issue, and most brains will auto-correct anyway.

KeithWiles:
Merged with last in series.

PR #38, opened 2022-05-05
build:update the install instructions for libbpf

KeithWiles:
Update the libbpf instructions when building from source or using
apt-get to install libbpf-dev package.

- CNDP needs libbpf version >=0.3.0 and <=0.6.1 to build.
- Make sure the install instructions give the correct build process for
  building v0.5.0 or v0.6.1.
- Updated the ansible main.yml file to pull and install the correct version
   of libbpf source code.
- Update the Dockerfile to pull and build v0.5.0.
- Update the linux_gsg guide to be inline with the install text.
- Change meson.build to require a libbpf version >=0.3.0 and <=0.6.1

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -7,16 +7,9 @@ instructions to run CNDP applications. For more information, refer to the CNDP d
 
 ## Prerequisites
 
-### apt proxy
-
-If required, create a proxy.conf and configure the apt proxy settings.
-
-```bash
-cat << EOF | sudo tee -a /etc/apt/apt.conf.d/proxy.conf
-Acquire::http::Proxy "http://user:password@proxy.server:port/";
-Acquire::https::Proxy "https://user:password@proxy.server:port/";
-EOF
-```
+If behind a proxy server you may need to setup a number of configurations to allow access via the server.
+Some commands i.e. apt-get, git, ssh, curl, wget and others will need configuration to work correctly.
+Please refer to apt-get, git and other documentations to enable access through a proxy server.
 
 ### dependencies
 
@@ -40,7 +33,15 @@ The [libbpf](https://github.com/libbpf/libbpf) is a dependency of CNDP. Starting
 it can be installed using apt. For earlier Ubuntu versions, or for users who want the latest code,
 it can be installed from source.
 
-### Install libbpf from package manager
+#### _Note:_
+
+Newer versions of libbpf greater than or equal to v0.7.0 require _libxdp_ to be installed. For now we
+can checkout a previous version v0.5.0 or v0.6.1 instead of installing _libxdp_.
+
+### Install libbpf-dev from package manager
+
+Use the following command on Ubuntu 20.10 and later to install the headers and libraries to build
+and run CNDP applications. If using an earlier Ubuntu version, you need to build libbpf from source.
 
 ```bash
 sudo apt-get install -y libbpf-dev
@@ -49,20 +50,24 @@ sudo apt-get install -y libbpf-dev
 ### Install libbpf from source
 
 ```bash
-https_proxy="https://user:password@proxy.server:port" git clone https://github.com/libbpf/libbpf.git
+git clone https://github.com/libbpf/libbpf.git
 cd libbpf
-git checkout ...
-make -C src
-sudo make -C src install
+git checkout v0.5.0  # or v0.6.1 if needing a newer version
+make -C src && sudo make -C src install
 ```
 
 The library and pkgconfig file is installed to /usr/lib64, which is not where the loader or
-pkg-config look. Fix this by editing the ldconfig and exporting the PKG_CONFIG_PATH.
+pkg-config looks. Fix this by editing the ldconfig file as suggested below.
 
 ```bash
 sudo vim /etc/ld.so.conf.d/x86_64-linux-gnu.conf
-# add /usr/lib64 line to the bottom of the file, save and exit.
+# add a line with /usr/lib64 to the bottom of the file, save and exit.
 sudo ldconfig
+```
+
+The following statement may be necessary if libbpf is installed from source instead of the package manager.
+
+```bash
 export PKG_CONFIG_PATH=/usr/lib64/pkgconfig
 ```
 
--- a/ansible/cndp-ansible/roles/install_libbpf/tasks/main.yml
+++ b/ansible/cndp-ansible/roles/install_libbpf/tasks/main.yml
@@ -27,7 +27,7 @@
 
 - name: Unarchive libbpf
   unarchive:
-    src: https://github.com/libbpf/libbpf/archive/refs/tags/v0.3.tar.gz
+    src: https://github.com/libbpf/libbpf/archive/refs/tags/v0.5.0.tar.gz
     dest: /tmp
     remote_src: yes
   when: bpf_static_installed.stat.exists == False or bpf_shared_installed.stat.exists == False
@@ -37,7 +37,7 @@
 
 - name: Install libbpf
   make:
-    chdir: /tmp/libbpf-0.3/src
+    chdir: /tmp/libbpf-0.5.0/src
     target: install
     params:
       NUM_THREADS: "{{ num_threads }}"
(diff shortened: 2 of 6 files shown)


INSTALL.md
```diff
@@ -49,20 +60,24 @@ sudo apt-get install -y libbpf-dev
 ### Install libbpf from source
 
 ```bash
-https_proxy="https://user:password@proxy.server:port" git clone https://github.com/libbpf/libbpf.git
+git clone https://github.com/libbpf/libbpf.git
```
jeffreybshaw, line 63: Should create a new section at the top of the file to mention the https_proxy is required if you are downloading from the internet. If you're behind a proxy, most people know they need to configure the proxy, so this would just be for completeness.

INSTALL.md
```diff
@@ -49,20 +60,24 @@ sudo apt-get install -y libbpf-dev
...
 cd libbpf
-git checkout ...
-make -C src
-sudo make -C src install
+git checkout v0.6.1
```
jeffreybshaw, line 65: If we used 0.5.0 we can align with at least the Ubuntu 22.04 package manager's version. I suppose its ok to recommend the newer version; personally I just use the package manager's version.

jeffreybshaw (changes requested):
The commit message is not quite accurate since you are also updating the Ansible build scripts.
If you are going to update the install instructions *and* ansible in the same commit, you might as well update the Dockerfile and linux_gsg as well. Otherwise I would prefer to see separate commits for each.

INSTALL.md
```diff
@@ -40,7 +40,19 @@ The [libbpf](https://github.com/libbpf/libbpf) is a dependency of CNDP. Starting
...
+#### _Note:_
+
+```bash
+Newer versions of libbpf greater than or equal to v0.7.0 require _libxdp_ to be installed. For now we
+can checkout a previous version v0.5.0 or v0.6.1 instead of installing _libxdp_.
+```
+
+### Install libbpf-dev from package manager
+
+Installing **libbpf-dev** installs the header files required to build CNDP.
```
jeffreybshaw, line 52: This sentence and the next one are redundant. This whole paragraph is oddly worded, and the line length is long.

```
Use on the following command on Ubuntu 20.10 and later to install the headers and libraries to build
and run CNDP applications. If using an earlier Ubuntu version, you need to build libbpf from source.
```

INSTALL.md
```diff
@@ -49,20 +61,24 @@ sudo apt-get install -y libbpf-dev
...
+pkg-config looks. Fix this by editing the ldconfig file as suggested below.
 
 ```bash
 sudo vim /etc/ld.so.conf.d/x86_64-linux-gnu.conf
-# add /usr/lib64 line to the bottom of the file, save and exit.
+# add a line with /usr/lib64 to the bottom of the file, save and exit.
 sudo ldconfig
+```
+
+The following export statment may not be required, but it does not hurt.
```
jeffreybshaw, line 79: typo => s/statment/statement

Suggest:
```
The following statement may be necessary if libbpf is installed from source instead of the package manager.
```

ansible/cndp-ansible/roles/install_libbpf/tasks/main.yml
```diff
@@ -27,7 +27,7 @@
 
 - name: Unarchive libbpf
   unarchive:
-    src: https://github.com/libbpf/libbpf/archive/refs/tags/v0.3.tar.gz
+    src: https://github.com/libbpf/libbpf/archive/refs/tags/v0.6.1.tar.gz
```
jeffreybshaw, line 30: Any reason to use 0.6.1 here but 0.5.0 everywhere else?

doc/guides/linux_gsg/linux_gsg.rst
```diff
@@ -145,7 +145,16 @@ On a NUMA machine with two nodes, pages should be allocated explicitly on separa
    For 1GB pages, it is not possible to reserve the hugepage memory after the system has booted.
 
 Prerequisites
-~~~~~~~~~~~~~~
+~~~~~~~~~~~~~
+
+Building CNDP from behind a proxy server
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+If building behind a proxy server the following is required before excuting a git clone or other commands needs to pass thru the proxy server.
```
jeffreybshaw, line 153: Line is long, I prefer to keep lines, even in documentation <100 chars for consistency with code.

couple of typos

s/excuting/executing
s/thru/through

"or other commands needs to pass thru the proxy server" should be "needing", not "needs"

INSTALL.md
```diff
@@ -40,7 +40,19 @@ The [libbpf](https://github.com/libbpf/libbpf) is a dependency of CNDP. Starting
 it can be installed using apt. For earlier Ubuntu versions, or for users who want the latest code,
 it can be installed from source.
 
-### Install libbpf from package manager
+#### _Note:_
+
+```bash
+Newer versions of libbpf greater than or equal to v0.7.0 require _libxdp_ to be installed. For now we
```
elzamath, line 46: This text is not a shell command so we don't need to enclose it within ```bash 

Italicizing libxdp didn't seem to work though: https://github.com/KeithWiles/cndp/blob/install_libbpf_update/INSTALL.md#note maybe because it's within the ``` bash block

doc/guides/linux_gsg/linux_gsg.rst
```diff
@@ -145,7 +145,16 @@ On a NUMA machine with two nodes, pages should be allocated explicitly on separa
...
+
+.. code-block:: console
+
+   https_proxy="password@proxy.server:port>https://user:password@proxy.server:port" git clone https://github.com/libbpf/libbpf.git
```
jeffreybshaw, line 157: This isn't right. Should be
```
export http_proxy="http://proxy.server:port"
export https_proxy="https://proxy.server:port"
```

And you can drop the "git clone" part which is irrelevant.

jeffreybshaw (changes requested):

elzamath (commented):
Do we want to edit the Github action to use libbpf v0.5.0 as well in this PR or a separate one? (https://github.com/CloudNativeDataPlane/cndp/blob/main/.github/workflows/smoke.yml#L28)

INSTALL.md
KeithWiles: I am starting to think this proxy statement is not required, if the developer is behind a proxy they had better know how to use it. We should not be trying to teach them how to setup a proxy.

doc/guides/linux_gsg/linux_gsg.rst
KeithWiles: We need to drop the proxy setup as the developer should already know how to setup his proxy and we should not be trying to teach them how to setup a proxy.

Just extra text and most likely will not apply to every possible proxy setup.

doc/guides/linux_gsg/linux_gsg.rst
```diff
@@ -145,35 +145,28 @@ On a NUMA machine with two nodes, pages should be allocated explicitly on separa
...
+Build libbpf
+~~~~~~~~~~~~
 
 The `libbpf <https://github.com/libbpf/libbpf>`_ is a dependency of CNDP. Starting with Ubuntu 20.10
-it can be installed using apt. For earlier Ubuntu versions, or for users who want the latest code,
-it can be installed from source.
+the libbpf libraries can be installed using apt-get. For earlier Ubuntu versions, or for users who
+want the latest code, it can be installed from source.
 
-**Install using apt**
```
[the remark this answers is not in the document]
KeithWiles: We were using a mixture of apt or apt-get so I picked one.

elzamath (approved):

INSTALL.md
jeffreybshaw: I agree.

doc/guides/linux_gsg/linux_gsg.rst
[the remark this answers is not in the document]
jeffreybshaw: Also, sometimes "apt" warns users that it does not have a stable interface, and recommends apt-get.

jeffreybshaw (approved):

PR #40, opened 2022-05-05
fix heap buffer overflow

elzamath:
By fuzzing mempool_create(), a heap buffer overflow was detected in
ring[] while in the ENQUEUE_PTRS macro of the __cne_ring_do_enqueue function.
This was because the actual memory that was allocated for the ring was only the size
of the struct cne_ring. This happened because of an integer overflow
error in the calculation of ring size in cne_ring_get_memsize_elem. The multiplication
of count * esize was evaluating to zero.

Typecasting the unsigned int count to be ssize_t ensured that the multiplication
of count and esize did not evaluate to zero (because of the integer overflow), the
correct ring size memory was allocated and the heap buffer overflow was fixed.

### DIFF
--- a/lib/core/ring/cne_ring.c
+++ b/lib/core/ring/cne_ring.c
@@ -46,7 +46,7 @@ cne_ring_get_memsize_elem(unsigned int esize, unsigned int count)
             "Requested number of elements is invalid, must be power of 2, and not exceed %u\n",
             CNE_RING_SZ_MASK);
 
-    sz = sizeof(struct cne_ring) + count * esize;
+    sz = sizeof(struct cne_ring) + (ssize_t)count * esize;
     sz = CNE_ALIGN(sz, CNE_CACHE_LINE_SIZE);
     return sz;
 }


KeithWiles:
Not sure I understand the reason for this change.

Did you fuzz the function cne_ring_get_memsize_elem() by itself?
The function returns size_t (doing esize * count) should not overflow a size_t 64 bit value. It may go negative it assigned to a signed variable.

elzamath:
> Did you fuzz the function cne_ring_get_memsize_elem() by itself? 
No, I fuzzed the mempool_create function actually using this fuzz target https://github.com/CloudNativeDataPlane/cndp/blob/main/test/fuzz/mempool_create.cc. 

The error reported was this:
image
When stepping through gdb, in this line in https://github.com/CloudNativeDataPlane/cndp/pull/40/files#diff-0dbb7cd9a7f3ec5e1b7d24be78dc4d02885da115e8b64aa40f38e5eac6ffd82eR49 the multiplication of count * esize was evaluating to zero . So, sz was equal to just sizeof(struct cne_ring). So, in __cne_ring_do_enqueue, ENQUEUE_PTRS macro the ring was being indexed beyond what was allocated (384 bytes = sizeof(struct cne_ring).

KeithWiles (approved):
LGTM

PR #45, opened 2022-05-09
Rust Loopback test: Fix send packets in loop and free packets not sent.

manojgop:
- Try to send packet in a loop with threshold on max tries.
 - Free packets which are not sent.


KeithWiles (approved):
LGTM but I do not know Rust very well :-(

PR #47, opened 2022-05-09
docker: base fedora build

maryamtahhan:
Add a Docker build for fedora images.


Makefile
```diff
@@ -70,7 +70,11 @@ snyk: FORCE
 
 docker-image: FORCE
 	docker build -t cndp --build-arg http_proxy=${http_proxy} \
-  --build-arg https_proxy=${http_proxy} -f containerization/docker/Dockerfile .
+  --build-arg https_proxy=${http_proxy} -f containerization/docker/ubuntu/Dockerfile .
+
+docker-fed-image: FORCE
```
maryamtahhan, line 75: I could look into creating a bash script similar to cne_build.sh to streamline the docker builds in the future to avoid and explosion of docker-**_os_**-image and docker-**_os_**-run rules in the makefiles... what do we think?

containerization/docker/fedora/Dockerfile
```diff
@@ -0,0 +1,41 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2020-2022 Intel Corporation.
+
+FROM fedora:35 AS build
```
maryamtahhan, line 4: not sure if we want the multistage build here - but could add the 2nd half similar to the ubuntu build if needed

jeffreybshaw: I suspect the size is probably similar between the build and runtime container images since the base image will most likely dominate the overall image size.

For development purposes, and as an example Dockerfile, this is good. Actually I prefer having the single stage for development since I can debug there too; so I'd almost prefer if the Ubuntu dockerfile just had a single stage. Not too big a deal, at least for me, though.

containerization/docker/fedora/Dockerfile
```diff
@@ -0,0 +1,41 @@
+# SPDX-License-Identifier: BSD-3-Clause
```
jeffreybshaw, line 1: Something we should do in a future PR is to create a build matrix for the "docker" workflow (.github/workflows/docker.yml) so we'll know if the container build is broken for Fedora. It also has the added benefit of testing compilation on Fedora.

jeffreybshaw (approved):

elzamath (approved):
Hi Maryam :) looks good to me, thanks!

PR #59, opened 2022-05-19
change punt node from TUN to Raw socket

KeithWiles:
The original code used a TUN interface to sent/recv traffic with the kernel
stack. This turned out to not work in all cases as a TUN is a point-2-point
type interface and sending traffic from CNDP to Linux stack would not get
sent to a host application waiting for packets not handled by CNET.

When redirecting say all UDP packets to CNDP to a destination port not handled
by CNDP these should be punted to the Linux stack for processing. This allows host
applications waiting on the destination port to process the packets. If not it should
discard the packets.

Let say we have CNET stack listening for UDP destination port 1234 and a packet arrives
for say dest-port 6666 then these packets must me punted to the Linux stack. Using
ethtool all UDP frames are redirected to CNDP. In the case of dest-port 6666,
we punt the packets to the linux kernel stack via a raw socket to handle.

Here is one way to setup for testing on two machines one running CNDP(DUT) and the
other one a normal machine. Start up cnet-graph with listening to 1234 or some port,
then using socat on both machines we can do this:

DUT:
$ socat -b 1400 - UDP-RECV:6666

Remote:
$ xxd /boot/vmlinuz > testfile.txt
$ cat testfile.txt | socat UDP-SENDTO:198.18.2.2:6666 -

On the DUT you should see the testfile.txt printed to the screen. Note the -b 1400 is
to force Linux to not support jumbo frames as af_xdp does not support it.

### DIFF
--- a/lib/cnet/punt/kernel_recv.c
+++ b/lib/cnet/punt/kernel_recv.c
@@ -39,6 +39,33 @@
 #include "kernel_recv_priv.h"
 #include "tun_alloc.h"
 
+static inline pktmbuf_t *
+alloc_rx_mbuf(kernel_recv_info_t *rx)
+{
+    if (rx->idx >= rx->cnt) {
+        uint16_t cnt;
+
+        rx->idx = 0; /* Reset the index value to start of rx_bufs array */
+        rx->cnt = 0;
+
+        cnt = pktmbuf_alloc_bulk(rx->pi, rx->rx_bufs, KERN_RECV_CACHE_COUNT);
+        if (cnt <= 0)
+            return NULL;
+
+        rx->cnt = cnt;
+    }
+
+    return rx->rx_bufs[rx->idx++];
+}
+
+static inline void
+free_rx_mbuf(kernel_recv_info_t *rx)
+{
+    if (rx->idx <= 0)
+        CNE_RET("rx_mbuf array is empty\n");
+    rx->idx--;
+}
+
 static inline void
 mbuf_update(pktmbuf_t **mbufs, uint16_t nb_pkts)
 {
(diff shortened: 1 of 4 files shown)


lib/cnet/punt/punt_kernel.c
```diff
@@ -37,40 +38,47 @@
...
 static __cne_always_inline void
 punt_kernel_process_mbuf(struct cne_node *node, pktmbuf_t **mbufs, uint16_t cnt)
 {
     punt_kernel_node_ctx_t *ctx = (punt_kernel_node_ctx_t *)node->ctx;
-    struct iovec v[cnt];
-    int fd;
 
-    if ((fd = tun_get_fd(ctx->tinfo)) >= 0) {
+    if (ctx->sock >= 0) {
+        struct cne_ipv4_hdr *ip4;
```
jeffreybshaw, line 54: we can add ipv6 later

lib/cnet/punt/punt_kernel.c
```diff
@@ -37,40 +38,47 @@
...
+            ip4 = pktmbuf_mtod(mbufs[i], struct cne_ipv4_hdr *);
+            len = pktmbuf_data_len(mbufs[i]);
+            buf = (char *)ip4;
+
+            sin.sin_family      = AF_INET;
+            sin.sin_port        = 0;
+            sin.sin_addr.s_addr = ip4->dst_addr;
+
+            if (sendto(ctx->sock, buf, len, 0, (struct sockaddr *)&sin, sizeof(sin)) < 0)
+                CNE_WARN("Unable to send packets: %s\n", strerror(errno));
```
jeffreybshaw, line 69: This log message could be overkill under high load. Ok for now, but eventually we'll want a stat increment or something rather than a log.

lib/cnet/punt/kernel_recv.c
```diff
@@ -200,40 +231,59 @@ kernel_recv_node_init(const struct cne_graph *graph __cne_unused, struct cne_nod
...
-    tun_dump(NULL, ctx->tinfo);
     return 0;
 }
 
+static void
+kernel_recv_node_fini(const struct cne_graph *graph __cne_unused, struct cne_node *node)
+{
+    kernel_recv_node_ctx_t *ctx = (kernel_recv_node_ctx_t *)node->ctx;
+
+    close(ctx->sock);
```
jeffreybshaw, line 270: I looked through the graph code, and the de-constructor is only called if the (all the nodes') constructor succeeded, when the graph is destroyed. This means that ctx->sock will reference a valid socket.

But there is a different problem, where if we have some nodes that fail to initialize later (after this one) we will never actually call any of the de constructors, so we'll never close this or free memory, etc. We should fix that at some point (different PR).

jeffreybshaw (approved):

lib/cnet/punt/punt_kernel.c
KeithWiles: We can create a punt6_kernel.c  version later, this way we do not need to worry about combining them into one.

lib/cnet/punt/punt_kernel.c
KeithWiles: I agree, it needs to be a stats counter.

lib/cnet/punt/kernel_recv.c
KeithWiles: Yes, I agree. I should have tested for ctx->sock >= 0 ??

jeffreybshaw: If we thought `fini` were called even when `init` fails, yes, but that isn't how the graph code works. It only calls `fini` functions when all the `init` functions in the graph succeed. Therefore we know `ctx->sock` will be valid, otherwise it would have failed in `init` and the `fini` functions would never be called (which is a different memory/resource leak problem)

PR #62, opened 2022-05-23
tcp:add tcp support to CNET

KeithWiles:
Update and rework the TCP support for CNET using a set of graph nodes to
manage the TCP protocol. The TCP protocol support can be enabled via the
meson_options.txt file.

Testing the TCP protocol using wireguard and other applications i.e wireshark, socat, telnet, ...
to verify the protocol and data managment is correct.

In this change a few examples needed to be update to support TCP and the changes to
the channel APIs. One change to the channel API is when the application is called
back to process Rx packets is the callback does not pass mbufs back in the call.
The packets are now placed on a receive queue and a call to chnl_recv() needs to
be done to get packets.

Note:
TCP active opens are not test yet. This PR is for anyone wanting to understand the changes or would like to test the code.
It does take a bit of effort to test, so please ask any questions if you want to test the code.

**_Also this is one big PR and I will break this down into smaller PR's when I get closer to submitting the changes._** 


lib/include/cne_vec.h
```diff
@@ -342,10 +342,29 @@ _vec_find_index(void **vec, void *v)
...
+ * @param _v
+ *   The pointer to the vector to remove the entries from
+ * @param _nb
+ *   The number of items to remove from the vector
+ */
+#define vec_remove(_v, _nb)                             \
+    do {                                                \
+        vec_hdr_t *h = vec_header(_v);                  \
+        int _n       = h->len - _nb;                    \
+        if (_n) {                                       \
```
jeffreybshaw, line 363: check `if (_n > 0)`

jeffreybshaw (changes requested):

lib/include/cne_vec.h
KeithWiles: updated.

lib/core/pktmbuf/pktmbuf.h
```diff
@@ -73,7 +73,8 @@ typedef struct pktmbuf_info_s {
     uint32_t bufcnt;                  /**< Number of buffers created */
     uint32_t bufsz;                   /**< Size of each buffer */
     uint32_t cache_sz;                /**< Cache size if needed for allocation cache */
-    mmap_type_t mtype;                /**< Type of MMAP memory to allocate */
+    uint32_t allocated;
```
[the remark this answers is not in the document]
KeithWiles: Turns out that field is not used, so removed.

KeithWiles:
Going to close this PR and reopen it after it is broken up into PRs for non-TCP issue.

PR #63, opened 2022-05-25
pmd memif fix overwriting mbuf header.

KeithWiles:
This change came from DPDK and the following commit message.

   "The 'dst_off' was reset in multi segment case.
   This caused memif buffer segment to write to
   beginning of mbuf, overwriting previous data.
   Fix it with this patch.

   Fixes: 09c7e63a71f9 ("net/memif: introduce memory interface PMD")
   Cc: stable@dpdk.org

   Reported-by: Ferruh Yigit <ferruh.yigit@xilinx.com>
   Signed-off-by: Joyce Kong <joyce.kong@arm.com>
   Reviewed-by: Ruifeng Wang <ruifeng.wang@arm.com>"


KeithWiles:
Just noticed this fix in DPDK and applied it here.

PR #69, opened 2022-05-26
fix warning in meson with run_command arguments

KeithWiles:
The run_command will be changed to require check: true|false in the future.

```
WARNING: You should add the boolean check kwarg to the run_command call.
         It currently defaults to false,
         but it will default to true in future releases of meson.
         See also: https://github.com/mesonbuild/meson/issues/9300

```

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -5,7 +5,7 @@ project('CNDP', 'C', 'cpp',
     # Get version number from file.
     # Fallback to "more" for Windows compatibility.
     version: run_command(find_program('cat', 'more'),
-        files('VERSION')).stdout().strip(),
+        files('VERSION'), check: true).stdout().strip(),
     license: 'BSD',
     default_options: [
         'buildtype=release',
@@ -231,7 +231,7 @@ endforeach
 compile_time_cpuflags = []
 
 # get binutils version for the workaround of Bug 97
-ldver = run_command('ld', '-v').stdout().strip()
+ldver = run_command('ld', '-v', check: true).stdout().strip()
 if ldver.contains('2.30') and cc.has_argument('-mno-avx512f')
     machine_args += '-mno-avx512f'
     message('Binutils 2.30 detected, disabling AVX512 support as workaround for bug #97')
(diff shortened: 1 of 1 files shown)


jeffreybshaw (approved):

meson.build
```diff
@@ -5,7 +5,7 @@ project('CNDP', 'C', 'cpp',
     # Get version number from file.
     # Fallback to "more" for Windows compatibility.
     version: run_command(find_program('cat', 'more'),
-        files('VERSION')).stdout().strip(),
+        files('VERSION'), check: true).stdout().strip(),
```
eli-schwartz, line 8: If you (are able to) bump your minimum version of Meson to 0.57.0 then you can do:

```meson
version: files('VERSION')
```
and meson will handle everything here for you. No need to try to portably find some kind of `cat` and run it as a command.

meson.build
```diff
@@ -231,7 +231,7 @@ endforeach
 compile_time_cpuflags = []
 
 # get binutils version for the workaround of Bug 97
-ldver = run_command('ld', '-v').stdout().strip()
+ldver = run_command('ld', '-v', check: true).stdout().strip()
```
eli-schwartz, line 234: This is fragile, because CC_LD may be gold or lld or mold and thus binutils versions aren't relevant. That being said, `find_program()` has a version kwarg and you can use it to look for specific versions -- if it isn't found, then ld probably exists but is a different version.

I added a `.version()` method in 0.62.0 (I know, very new :() so you can do `ldver = find_program('ld').version()`.

meson.build
KeithWiles: We are trying to support Ubuntu 20.04, 21.04 and 22.04 LTS versions. We will remove 20.04 support soon.

The standard installed package versions for meson for Ubuntu versions are:
20.04 - 0.53.2
21.04 - 0.57.0
22.04 - 0.61.2

I would like to bump the minimum version but the latest version we can safely bump to is 21.04 0.57.0.
I do not think we want to force developers to install different versions then what comes with the distro release.

When we drop support for 20.04 LTS we can move to 0.57.0 and fix this one.

meson.build
KeithWiles: The newer .version() support would be great, but we are limited to older versions for now.

KeithWiles: Thank you for the review, if we can find a safe way to support the latest meson/ninja I would be happy to do it.

PR #72, opened 2022-06-02
testcne: update meson with metrics test

sushmasi:

### DIFF
--- a/test/testcne/meson.build
+++ b/test/testcne/meson.build
@@ -128,6 +128,7 @@ test_names = [
     'log',
     'mbuf',
     'mempool',
+    'metrics',
     'mmap',
     'pkt',
     'pktcopy',


jeffreybshaw:
This is failing because we do not run the tests as root, which is required to open the uds socket used by the metrics library.
One thing we could try is to create a new group of tests, i.e. "sudo_test_names" that run with "sudo meson test -C builddir --suite su-short" where we run the tests that require sudo.

KeithWiles (approved):

test/testcne/meson.build
```diff
@@ -128,6 +128,7 @@ test_names = [
     'log',
     'mbuf',
     'mempool',
+    'metrics',
```
jeffreybshaw, line 131: Please do not do this. We don't run the meson tests as root, and this test willl fail if it is not run as root. See #75

jeffreybshaw (changes requested):

jeffreybshaw:
@sushmasi - can you close this PR?

sushmasi:
Changes taken care by commit 8770080db17e8df576b33622651b2ef985fef5f4

PR #73, opened 2022-06-02
Unit test fixes

jeffreybshaw:
* Fix bug in lport group umem setup
* Fix bug in jcfg test that exits early if a file cannot be parsed
* Fix pktcpy test and disable color output when using meson test since it writes to a file, not a terminal
* Update the GitHub Smoke test action to upload the resulting testlog.txt to help debug errors

### DIFF
--- a/.github/workflows/smoke.yml
+++ b/.github/workflows/smoke.yml
@@ -35,3 +35,9 @@ jobs:
 
     - name: Short test
       run: meson test -C builddir --suite short
+
+    - name: Upload testlog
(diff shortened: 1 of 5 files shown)


KeithWiles (approved):
LGTM

jeffreybshaw:
Smoke test isn't running since upload-artifact isn't allowed in the allowed actions list. I updated that, but there's no way to re-run a failed job that failed because of a disallowed action. I'm going to close and re-open to see if the CI runs.

PR #75, opened 2022-06-03
Testcne skipped

jeffreybshaw:
* add ability to mark tests as "skipped". This is useful to denote tests that don't pass or fail, they simply cannot be run. This is helpful when running `meson test` since the result will still be `0` if no test fails
* skip the metrics test if we run into a permissions error, which is not really a failure of the test, but rather a failure of running without adequate permission
* add some missing tests to the `meson test` command

Thanks @eli-schwartz for the tip about skipping tests.

### DIFF
--- a/test/common/tst_info.c
+++ b/test/common/tst_info.c
@@ -16,10 +16,21 @@
 struct tst_stat {
     atomic_uint fail;
     atomic_uint pass;
+    atomic_uint skip;
 };
 
 /* Global test statistics updated by tst_end(). */
 static struct tst_stat tst_stats;
+
+int
+tst_exit_code(void)
+{
+    if (atomic_load(&tst_stats.fail))
+        return EXIT_FAILURE;
+    else if (atomic_load(&tst_stats.skip))
+        return EXIT_SKIPPED;
+    return EXIT_SUCCESS;
+}
 
 uint32_t
 tst_summary(void)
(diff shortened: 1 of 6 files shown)


test/common/tst_info.c
```diff
@@ -16,11 +16,22 @@
...
 static struct tst_stat tst_stats;
 
+int
+tst_exit_code(void)
+{
+    if (atomic_load(&tst_stats.fail))
+        return EXIT_FAILURE;
+    else if (atomic_load(&tst_stats.skip))
+        return EXIT_SKIPPED;
+    return EXIT_SUCCESS;
```
KeithWiles, line 32: If running multiple tst_start()/tst_end() in a single test would the above prove wrong?

I do the following in the same test function called from main:
tst = tst_start("foobar test1"); 
tst_end(tst, SUCCESS);

tst = tst_start("foobar test2"); 
tst_end(tst, SKIP);

tst = tst_start("foobar test3"); 
tst_end(tst, SUCCESS);

The above would report test skipped, but two of the tests passed. Because tst_stats is global for all tests. What
I would expect to see the in above is two passed and one skipped being printed.

The metrics test would work as it only has one set of tst_start/tst_end in the single test.

KeithWiles (changes requested):

test/common/tst_info.c
KeithWiles: I assumed the test results would be printed for the log, but that does not matter in this question.

If a test reports skipped on exit, it is a bit mis-leading to me, but I can accept it.

eli-schwartz: Using TAP (rather than, or as an option instead of, "exitcode") for reporting test statuses would allow trivially emitting different statuses per subtest. So you could indeed preserve this information.

PR #77, opened 2022-06-08
mutex helper routines handle pthread APIs cleanly

KeithWiles:
The pthread creation of a mutex is difficult to get right and
have klocwork accept the many issues with error codes.

Create two routines to init and destroy mutex structures. The
helper routine allows the caller to set attributes in the mutex
i.e recursive type.


jeffreybshaw (changes requested):
Build is failing https://github.com/CloudNativeDataPlane/cndp/runs/6801205358?check_suite_focus=true#step:5:311

jeffreybshaw (approved):

PR #79, opened 2022-06-10
fix klocwork issues

KeithWiles:
ID-167 - Convert to use the new mutex routines.
ID 225 and 226 were also fixed to free 'attr' and use the helper routines.

### DIFF
--- a/lib/include/cne_mutex_helper.h
+++ b/lib/include/cne_mutex_helper.h
@@ -30,20 +30,36 @@ extern "C" {
 static inline int
 cne_mutex_create(pthread_mutex_t *mutex, int flags)
 {
-    int ret = EFAULT;
+    pthread_mutexattr_t attr;
+    int inited = 0, ret = EFAULT;
 
-    if (mutex) {
-        pthread_mutexattr_t attr;
+    if (!mutex)
+        goto err;
 
-        ret = pthread_mutexattr_init(&attr);
-        if (ret == 0) {
-            ret = pthread_mutexattr_settype(&attr, flags);
-            if (ret == 0) {
-                ret = pthread_mutex_init(mutex, &attr);
-                if (ret == 0 && pthread_mutexattr_destroy(&attr) == 0)
-                    return 0;
-            }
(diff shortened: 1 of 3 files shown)


lib/include/cne_mutex_helper.h
```diff
@@ -30,19 +30,35 @@ extern "C" {
...
+    __do(pthread_mutex_init(mutex, &attr));
+
+    __do(pthread_mutexattr_destroy(&attr));
+
+#undef __do
+
+    return 0;
+err:
+    if (inited) {
+        if (pthread_mutexattr_destroy(&attr)) {
```
jeffreybshaw, line 60: Why have an if statement here? I know you're trying to convince klocwork that you're checking the return value, but it seems like we should do something like log an error if this fails rather than check the return value just to ignore it.

lib/include/cne_mutex_helper.h
```diff
@@ -30,19 +30,35 @@ extern "C" {
...
+        goto err;
+
+#define __do(_exp)    \
+    do {              \
+        ret = _exp;   \
+        if (ret)      \
+            goto err; \
+    } while (0 /* CONSTCOND */)
+
+    __do(pthread_mutexattr_init(&attr));
```
jeffreybshaw, line 46: I don't like the custom macro definition just to check a few failure cases, but if you strongly prefer this way I'll defer to you.

lib/include/cne_mutex_helper.h
KeithWiles: I can print out a message, but I did not want to lose the root failure.

lib/include/cne_mutex_helper.h
KeithWiles: In my head it seems cleanup and simpler. I know what the macro does and it is done the same way everywhere. I am going to keep it.

jeffreybshaw (approved):
BTW- listing the klocwork IDs in teh commit message isn't necessary as these are internal references.

PR #80, opened 2022-06-10
testcne: fix klocwork issue by freeing buffer pointer

sushmasi:

### DIFF
--- a/test/testcne/metrics_test.c
+++ b/test/testcne/metrics_test.c
@@ -75,6 +75,9 @@ metrics_test(void)
 
     tst_ok("PASS --- TEST: Added lport stats to metrics buffer\n");
 
+    free(client->buffer);
+    client->buffer = NULL;
+
     ret = metrics_destroy();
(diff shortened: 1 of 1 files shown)


test/testcne/metrics_test.c
```diff
@@ -75,6 +75,9 @@ metrics_test(void)
 
     tst_ok("PASS --- TEST: Added lport stats to metrics buffer\n");
 
+    free(client->buffer);
```
jeffreybshaw, line 78: I think this should fix the klocwork issue.

jeffreybshaw (approved):

test/testcne/metrics_test.c
KeithWiles: I think it will fix the Klocwork issue.

The only concern is it exposes the client->buffer pointer as something the developer needs to know to free. It may have been better to create a set of destroy functions in UDS and Metrics, but for now it will work. The other solution is a bit more involved.

PR #86, opened 2022-06-15
Fix jcfg lport groups

jeffreybshaw:
jcfg: fix lport group mapping onto same thread

The lport group processing fails for two groups mapping to a single
thread. This is because the lport list size was not updated after
reallocating space for the new lports in the first lport group. The
second lport group saw the old size, so reallocated a smaller array than
is necessary to store the existing plus new lports in the per-thread
array.

jcfg: fix lport group umem creation

A default umem is created for all logical ports belonging to a logical
port group which does not otherwise define a umem. Previously, a umem
was created for each lport group, not a single one for all lports. To
avoid a packet copy when transmitting packets received on a lport in a
different group, all the groups should use a common umem.


KeithWiles (approved):
LGTM

PR #87, opened 2022-06-16
cnet:general cleanup and remove useless code

KeithWiles:
A number of small changes to reduce the number of changes in the next
CNET TCP updates.

- Add some meson options to set the max number or PCB, TCB and channel entries.
- In a number of locations the constructor routine would register both create/destroy
  function with cnet_add_instance() removed the routines that did nothing.
  - Other files cnet_rtshow.c, cnet_stk.c, cnet_upd_chnl.c, tcp_output.c
- Changes in files like cnet.c init routine was to simplfy the routine and make sure
  the things were cleaned up on error exit.
- The eth_rx.c node should skip L3 headers by adjusting mbuf offset before moving to
  the next node, which then means the IP input routine does not need to do the adjustment.
- The pcb routine to dump out details of the pcb was removed and the output routine
  now adds some color to the output information.
- A couple minor mis-spelling were corrected.

### DIFF
--- a/lib/cnet/ipv4/ip4_input.c
+++ b/lib/cnet/ipv4/ip4_input.c
@@ -106,21 +106,35 @@ ip4_input_node_process(struct cne_graph *graph, struct cne_node *node, void **ob
         dip[0] = dip[1] = dip[2] = dip[3] = 0;
 
         /* Extract DIP from mbufs plus validate the IP header checksum. */
-        ip4[0] = pktmbuf_adjust(mbuf0, struct cne_ipv4_hdr *, mbuf0->l2_len);
-        ip4[1] = pktmbuf_adjust(mbuf1, struct cne_ipv4_hdr *, mbuf1->l2_len);
-        ip4[2] = pktmbuf_adjust(mbuf2, struct cne_ipv4_hdr *, mbuf2->l2_len);
-        ip4[3] = pktmbuf_adjust(mbuf3, struct cne_ipv4_hdr *, mbuf3->l2_len);
+        ip4[0] = pktmbuf_mtod(mbuf0, struct cne_ipv4_hdr *);
+        ip4[1] = pktmbuf_mtod(mbuf1, struct cne_ipv4_hdr *);
+        ip4[2] = pktmbuf_mtod(mbuf2, struct cne_ipv4_hdr *);
+        ip4[3] = pktmbuf_mtod(mbuf3, struct cne_ipv4_hdr *);
 
-        if (likely(cne_ipv4_cksum(ip4[0]) == 0))
+        /* Adjust the data length for an IPv4 packet to the size given in the header. */
+        pktmbuf_data_len(mbuf0) = be16toh(ip4[0]->total_length);
+        pktmbuf_data_len(mbuf1) = be16toh(ip4[1]->total_length);
+        pktmbuf_data_len(mbuf2) = be16toh(ip4[2]->total_length);
+        pktmbuf_data_len(mbuf3) = be16toh(ip4[3]->total_length);
+
+        /*
+         * When the total length exceeds mbuf size, the size check/checksum below will
+         * detect the invalid size/packet which will be dropped as 'dip[n]' is zero.
+         */
+        if (likely(pktmbuf_data_len(mbuf0) < pktmbuf_buf_len(mbuf0)) &&
+            likely(cne_ipv4_cksum(ip4[0]) == 0))
             dip[0] = be32toh(ip4[0]->dst_addr);
 
-        if (likely(cne_ipv4_cksum(ip4[1]) == 0))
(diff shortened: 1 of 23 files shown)


lib/cnet/ipv4/ip4_input.c
```diff
@@ -106,10 +106,16 @@ ip4_input_node_process(struct cne_graph *graph, struct cne_node *node, void **ob
...
-        ip4[1] = pktmbuf_adjust(mbuf1, struct cne_ipv4_hdr *, mbuf1->l2_len);
-        ip4[2] = pktmbuf_adjust(mbuf2, struct cne_ipv4_hdr *, mbuf2->l2_len);
-        ip4[3] = pktmbuf_adjust(mbuf3, struct cne_ipv4_hdr *, mbuf3->l2_len);
+        ip4[0] = pktmbuf_mtod(mbuf0, struct cne_ipv4_hdr *);
+        ip4[1] = pktmbuf_mtod(mbuf1, struct cne_ipv4_hdr *);
+        ip4[2] = pktmbuf_mtod(mbuf2, struct cne_ipv4_hdr *);
+        ip4[3] = pktmbuf_mtod(mbuf3, struct cne_ipv4_hdr *);
+
+        /* Adjust the data length for an IPv4 packet to the size given in the header */
+        pktmbuf_data_len(mbuf0) = be16toh(ip4[0]->total_length);
```
jeffreybshaw, line 115: Should you trust the network packet here? I think we should at least verify that ip4.total_length is <= the maximum length available in the mbuf. For example, if the length were 0xFFFF, that would be bad.

lib/cnet/ipv4/ip4_output.c
```diff
@@ -63,24 +65,28 @@ ip4_header(pktmbuf_t *m, uint16_t nxt)
...
-    ip->src_addr        = md->laddr.cin_addr.s_addr;
-    ip->time_to_live    = pcb->ttl;
-    ip->type_of_service = pcb->tos;
-    ip->next_proto_id   = pcb->ip_proto;
+        return nxt;
+
     ip->version_ihl     = (IPv4_VERSION << 4) | (sizeof(struct cne_ipv4_hdr) / 4);
+    ip->type_of_service = pcb->tos;
     ip->total_length    = htobe16(pktmbuf_data_len(m));
+    ip->packet_id       = 0;
```
jeffreybshaw, line 75: not necessary since you're setting this to nif->ip_ident later anyway.

lib/cnet/pcb/cnet_pcb.c
```diff
@@ -203,29 +180,41 @@ pcb_show(mempool_t *mp __cne_unused, void *obj_cb_arg __cne_unused, void *obj,
...
 static void
-_pcb_dump(stk_t *stk, int type)
+__pcb_iterate(mempool_t *mp __cne_unused, void *obj_cb_arg __cne_unused, void *obj,
+              unsigned n __cne_unused)
+{
+    __pcb_show((struct pcb_entry *)obj);
+}
+
+static void
+_pcb_dump(stk_t *stk)
```
jeffreybshaw, line 194: Why do you have a leading underscore here?

lib/cnet/ipv4/ip4_input.c
KeithWiles: Added test for length to be valid.

lib/cnet/ipv4/ip4_output.c
KeithWiles: Removed.

lib/cnet/pcb/cnet_pcb.c
KeithWiles: Removed the underscores.

lib/cnet/ipv4/ip4_input.c
```diff
@@ -194,9 +208,12 @@ ip4_input_node_process(struct cne_graph *graph, struct cne_node *node, void **ob
         n_left_from -= 1;
 
         /* Extract DIP from mbuf */
-        ip4[0] = pktmbuf_adjust(mbuf0, struct cne_ipv4_hdr *, mbuf0->l2_len);
+        ip4[0] = pktmbuf_mtod(mbuf0, struct cne_ipv4_hdr *);
         dip[0] = be32toh(ip4[0]->dst_addr);
 
+        /* Adjust the data length for an IPv4 packet to the size given in the header */
+        pktmbuf_data_len(mbuf0) = be16toh(ip4[0]->total_length);
```
jeffreybshaw, line 215: Need the length check here too

jeffreybshaw (changes requested):

lib/cnet/ipv4/ip4_input.c
KeithWiles: Fixed, which it appears I forgot to do the checksum too.

jeffreybshaw (approved):

PR #92, opened 2022-06-21
Make Rust Cne Packet type to be safely used across threads.

manojgop:
- Add Send/Sync marker trait for Packet type.
- Refactor Packet APIs to be safely used across threads.


KeithWiles (approved):
It seems reasonable, but not an expert.

PR #93, opened 2022-06-21
add fuzz test for mmap_alloc

elzamath:

### DIFF
--- a/test/fuzz/mmap_alloc.cc
+++ b/test/fuzz/mmap_alloc.cc
@@ -1,0 +1,44 @@
+/* SPDX-License-Identifier: BSD-3-Clause
+ * Copyright (c) 2021-2022 Intel Corporation
+ */
+
+#include <cne_mmap.h>
+#include <string.h>
+#include <assert.h>
+/* Prototype required to fix "no previous prototype for function" error */
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size);
+
+/* When launching the mmap_alloc fuzz test, the -rss_limit_mb=N parameter
+ * needs to be passed. Otherwise, an out of memory error will happen.
+ */
+extern "C" int
+LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
+{
+    char *offset = NULL;
+    mmap_t *mm   = NULL;
+
+    struct {
+        uint32_t bufcnt      = 0;
+        uint32_t bufsz       = 0;
+        mmap_type_t hugepage = MMAP_HUGEPAGE_DEFAULT;
+    } args;
+
+    if (size < sizeof(args))
+        return 0;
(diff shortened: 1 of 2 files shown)


test/fuzz/mmap_alloc.cc
```diff
@@ -0,0 +1,29 @@
...
+ */
+extern "C" int
+LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
+{
+    uint32_t bufcnt, bufsz;
+    mmap_type_t hugepage;
+    mmap_t *mm;
+
+    if (size > sizeof(bufsz))
+        size = sizeof(bufsz);
```
KeithWiles, line 21: The bufsz variable is only used to calculate the value of size when the argument passed in was > 4. Change the sizeof(bufsz) to sizeof(uint32_t)

If size > 4 bytes (sizeof(uint32_t)) then size becomes 4. Can the size value be zero, if so then we can just leave the routine.

test/fuzz/mmap_alloc.cc
```diff
@@ -0,0 +1,29 @@
...
+    memcpy(&bufcnt, data, size);
+    hugepage = (mmap_type_t)(bufcnt % MMAP_HUGEPAGE_CNT);
```
KeithWiles, line 23: Bufcnt could be some random value if size == 0, so setting it to zero at the top of the function is reasonable. If size is less than 4 then we could have some random value in the remaining bytes of the bufcnt variable, the reason to set it to zero up front.

test/fuzz/mmap_alloc.cc
```diff
@@ -0,0 +1,29 @@
...
+    mm       = mmap_alloc(bufcnt, size, hugepage);
```
KeithWiles, line 24: Here if we pass in size == 0  or bufcnt == 0 then mm could == NULL and the mmap_addr_at_offset(mm, size) would return NULL. If mmap_alloc() returns NULL then we need to report an error, unless the arguments passed to the function would always return NULL as they are invalid.

test/fuzz/mmap_alloc.cc
```diff
@@ -0,0 +1,29 @@
...
+    mmap_addr_at_offset(mm, size);
```
KeithWiles, line 25: Should we check the return value of mmap_addr_at_offset() to be valid and within the range of the mapped memory? Also using size as the offset will always give us 0-4 bytes offset. Maybe better to use some value from 0 - (bufcnt * size) and in the code now size can never be larger than 4. I do not think this is what you had in mind maybe??

KeithWiles (changes requested):

KeithWiles:
Let's get on a conf call and discuss this one, I feel like I am missing something here.

Can we have a call today? If so please set one up for the afternoon. I believe the only meeting I have is 2-2:30 CST today.

KeithWiles (approved):
Looks good.

PR #94, opened 2022-06-23
move setlocale to a common location

KeithWiles:
To reduce the number of calls the developer needs to make in his
application we move the setlocale() call to the cne initialization
function.


jeffreybshaw (approved):
This is OK, unless the application wants a different locale setting, in which case it would need to know to reset it after a call to cne_init(), which is most likely not an issue at the moment.

PR #104, opened 2022-06-29
fix the function comments in cthread api lib

sushmasi:


KeithWiles (approved):
Thanks LGTM

PR #107, opened 2022-06-29
go-binding: Add New to allow creation with a Config struct

nwaples:
Not all configuration is stored in a single file. Adding the ability to create with a Config struct adds flexibility to use your own config store, embed the config in an existing config file, or even auto generate the config.


KeithWiles (approved):
Thanks, LGTM

Just need the sign off lines added.

PR #109, opened 2022-06-29
ring:add a cne_ring_init to use external memory

KeithWiles:
Add a new API cne_ring_init() to allow the caller to pass in the memory
to build the ring into for allowing the caller to supply the memory region.

The new API becomes useful when we need to share memory or the application needs
to build a ring into a already allocated memory region.

### DIFF
--- a/lib/core/ring/cne_ring.c
+++ b/lib/core/ring/cne_ring.c
@@ -30,10 +30,13 @@
 #define POWEROF2(x)       ((((x)-1) & (x)) == 0)
 #define RING_DFLT_ELEM_SZ sizeof(void *) /** The default ring element size*/
 
-size_t
+ssize_t
 cne_ring_get_memsize_elem(unsigned int esize, unsigned int count)
 {
     ssize_t sz;
+
+    if (esize == 0)
+        esize = RING_DFLT_ELEM_SZ;
 
     /* Check if element size is a multiple of 4B */
     if (esize % 4 != 0)
@@ -52,10 +55,10 @@ cne_ring_get_memsize_elem(unsigned int esize, unsigned int count)
 }
 
 /* return the size of memory occupied by a ring */
-size_t
+ssize_t
 cne_ring_get_memsize(unsigned int count)
 {
-    return cne_ring_get_memsize_elem(sizeof(void *), count);
+    return cne_ring_get_memsize_elem(RING_DFLT_ELEM_SZ, count);
 }
 
 void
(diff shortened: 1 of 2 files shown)


lib/core/ring/cne_ring.c
```diff
@@ -172,38 +176,61 @@ cne_ring_create(const char *name, unsigned int esize, unsigned int count, unsign
...
+        }
+        ring_mem = addr;
+        memset(ring_mem, 0, ring_size);
+    } else {
+        /* Reserve a memory region for this ring
+         * Calloc can return pointer which is not cache aligned.
+         * In that case cne_ring_setup calling memset will segfault on clang compiled
+         * with -O3. Tested with clang-10.0.0-4ubuntu1 and gcc-9.3.0-10ubuntu2
+         */
+        ring_mem = calloc(1, ring_size + CNE_CACHE_LINE_SIZE - 1);
```
jeffreybshaw, line 206: Instead of checking for ring_mem != NULL later, why not check it here and save yourself from needing an extra if statement later. Also, it seems odd to set flags |= RING_F_ALLOCATED if the memory allocation fails (i know free can take a NULL address, but still, it seems off).

lib/core/ring/cne_ring_api_internal.h
```diff
@@ -20,6 +20,8 @@
 extern "C" {
 #endif
 
+#define RING_F_ALLOCATED 0x8000 /**< The ring structure and data was allocated */
```
jeffreybshaw, line 23: I know this flag is only used for internal alloc/free, but should we move it to cne_ring_api.h where the other flags are defined? Somebody might accidentally add another flag and choose this same value without realizing its already in use. Also, we provide a cne_ring_get_flags() which would return the entire flags, which might confuse applications to see undefined flags set.

lib/core/ring/cne_ring.c
```diff
@@ -172,38 +176,61 @@ cne_ring_create(const char *name, unsigned int esize, unsigned int count, unsign
...
+        flags |= RING_F_ALLOCATED;
```
jeffreybshaw, line 207: We should verify that the user did not set any flags other than RING_F_SP_ENQ, RING_F_SC_DEQ, or RING_F_EXACT_SZ. If the user sets RING_F_ALLOCATED when passing their own memory, the application might crash if the memory is freed twice.

jeffreybshaw (changes requested):

lib/core/ring/cne_ring.c
KeithWiles: added the suggested change.

lib/core/ring/cne_ring.c
KeithWiles: Checked to make sure only valid flags are set by caller.

lib/core/ring/cne_ring_api_internal.h
KeithWiles: Originally I did have it in the cne_ring_api.h file :-( moved it back.

jeffreybshaw (commented):
Looks good, but I think we should have a testcne case that tests this.

jeffreybshaw (approved):

KeithWiles:
I can work on that for a different PR, would that be fine.

PR #110, opened 2022-06-29
add pktmbuf external metadata

KeithWiles:
The metadata was contained in the headroom of the pktmbuf and it is
limited in size to max headroom size normally 128 bytes.
The headroom size is also used for adding protcol headers and if the
room is consumed by the metadata then headers can not be added.

Created a new API pktmbuf_pool_cfg_create() to be able to create a
external metadata buffer array indexed by the pktmbuf_t.metadata_idx
value. The new metadata array of buffers must be a multiple of a
cacheline (64 bytes) and the number of buffers will match the number of
pktmbuf_t buffers in the mempool.

If the new API pktmbuf_pool_cfg_create() is not used the metadata buffer
will be part of the headroom in the pktmbuf buffer.


lib/core/pktmbuf/pktmbuf.c
```diff
@@ -104,38 +107,69 @@ pktmbuf_iterate(pktmbuf_info_t *pi, pktmbuf_cb_t cb, void *ud)
     return 0;
 }
 
+int
+pktmbuf_pool_cfg(pktmbuf_pool_cfg_t *c, void *addr, uint32_t bufcnt, uint32_t bufsz,
+                 uint32_t cache_sz, void *metadata, uint32_t metadata_bufsz, mbuf_ops_t *ops)
+{
+    if (!c || !addr || !bufcnt || !bufsz)
```
jeffreybshaw, line 114: This needs to change to allow `bufcnt` and `bufsz` to be 0.

You might as well do the same thing that is done in pktmbuf_pool_cfg_create()...

```
    CNE_MAX_SET(cache_sz, MEMPOOL_CACHE_MAX_SIZE);
    CNE_DEFAULT_SET(cache_sz, 0, MEMPOOL_CACHE_MAX_SIZE);
    CNE_DEFAULT_SET(bufcnt, 0, DEFAULT_MBUF_COUNT);
    CNE_DEFAULT_SET(bufsz, 0, DEFAULT_MBUF_SIZE);

    if (CNE_CACHE_LINE_ROUNDUP(bufsz) != bufsz)
        err!!!
```

lib/core/pktmbuf/pktmbuf.c
```diff
@@ -104,38 +107,69 @@ pktmbuf_iterate(pktmbuf_info_t *pi, pktmbuf_cb_t cb, void *ud)
...
+        return -1;
+
+    if ((!metadata && metadata_bufsz) || (metadata && !metadata_bufsz))
+        return -1;
+
+    /* Size of metadata buffer should be a multiple of a cacheline */
+    if (metadata && ((metadata_bufsz % CNE_CACHE_LINE_SIZE) != 0))
```
jeffreybshaw, line 121: Also, maybe re-use what's done above:

```
if (metadata && CNE_CACHE_LINE_ROUNDUP(metadata_bufsz) != metadata_bufsz)
```

jeffreybshaw (changes requested):

lib/core/pktmbuf/pktmbuf.c
```diff
@@ -121,21 +123,54 @@ pktmbuf_pool_create(char *addr, uint32_t bufcnt, uint32_t bufsz, uint32_t cache_
 
     /* Make sure the mbuf size is cache aligned */
     if (CNE_CACHE_LINE_ROUNDUP(bufsz) != bufsz)
+        CNE_ERR_RET("bufsz is not cache aligned\n");
+
+    if ((!metadata && metadata_bufsz) || (metadata && !metadata_bufsz))
+        CNE_ERR_RET("metadata and metadata_bufsz are invalid\n");
```
jeffreybshaw, line 129: nit: technically its "and/or".

lib/core/pktmbuf/pktmbuf.c
```diff
@@ -121,21 +123,54 @@ pktmbuf_pool_create(char *addr, uint32_t bufcnt, uint32_t bufsz, uint32_t cache_
...
+
+    /* Size of metadata buffer should be a multiple of a cacheline */
+    if (metadata && ((metadata_bufsz % CNE_CACHE_LINE_SIZE) != 0))
```
jeffreybshaw, line 132: Can we be consistent and use (CNE_CACHE_LINE_ROUNDUP(metadata_bufsz) != metadata_bufsz)? Or change the above line 128 to use this.

lib/core/pktmbuf/pktmbuf.c
KeithWiles: changed to and/or

lib/core/pktmbuf/pktmbuf.c
KeithWiles: Convert to use CNE_CACHE _LINE_ROUNDUP

jeffreybshaw (approved):

PR #121, opened 2022-07-13
Add Rust package keyword.

manojgop:
Add keywords in Rust cne and Rust cne-sys crates which will help
to search for these crates when crates are published in crates.io

### DIFF
--- a/lang/rs/apis/cne/Cargo.toml
+++ b/lang/rs/apis/cne/Cargo.toml
@@ -10,6 +10,7 @@ readme = "README.md"
 version = "0.1.0"
 edition = "2021"
 license = "BSD-3-Clause"
+keywords = ["networking", "cloud native", "AF_XDP"]
 
 [lib]
 name = "cne"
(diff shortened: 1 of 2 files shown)


lang/rs/apis/cne/Cargo.toml
```diff
@@ -10,6 +10,7 @@ readme = "README.md"
 version = "0.1.0"
 edition = "2021"
 license = "BSD-3-Clause"
+keywords = ["networking","cloud native", "AF_XDP"]
```
jeffreybshaw, line 13: nit: any reason not to put a space after the first comma? It's just inconsistent. We might as well get it right the first time.

lang/rs/bindings/cne-sys/Cargo.toml
```diff
@@ -10,6 +10,7 @@ readme = "README.md"
 version = "0.1.0"
 edition = "2021"
 license = "BSD-3-Clause"
+keywords = ["networking","cloud native", "AF_XDP"]
```
jeffreybshaw, line 13: space after first comma?

lang/rs/apis/cne/Cargo.toml
manojgop: Good catch. Cargo clippy didn't catch this :). I've updated PR

lang/rs/bindings/cne-sys/Cargo.toml
manojgop: Updated

jeffreybshaw (approved):

PR #123, opened 2022-07-14
go-bindings: add Pointer method to LocklessRing

nwaples:
Add Pointer method to expose the ring pointer. Allows
it to be used with CGO and have C threads communicate
with Go threads via rings.

Signed-off-by: Nicholas Waples <nwaples@gmail.com>


KeithWiles (approved):
LGTM

PR #129, opened 2022-07-16
add function to get information

KeithWiles:
Adding a function Info() to get the stats and some basic information
about the msgChan setup. Added a new function called Pointers() which
returns the Recv and Send ring pointers.

### DIFF
--- a/lang/go/bindings/cne/msgchan.go
+++ b/lang/go/bindings/cne/msgchan.go
@@ -24,6 +24,17 @@ type MsgChannel struct {
 	mChan unsafe.Pointer
 }
 
+type MsgChannelInfo struct {
+	RecvRing     unsafe.Pointer
+	SendRing     unsafe.Pointer
+	ChildCount   int
+	SendCalls    uint64
+	SendCnt      uint64
+	RecvCalls    uint64
+	RecvCnt      uint64
+	RecvTimeouts uint64
+}
+
 var msgChannel map[string]*MsgChannel
 var msgChanMu sync.Mutex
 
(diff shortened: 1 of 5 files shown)


lib/usr/clib/msgchan/msgchan.h
```diff
@@ -25,11 +25,28 @@
 extern "C" {
 #endif
 
+#ifndef MC_NAME_SIZE
```
jeffreybshaw, line 28: I'm assuming this is because the "priv" file needs this definition too? The duplication is not ideal, could the "priv" header just include this one? msgchan.c includes both anyway so there's no benefit for compilation time or anything.

lang/go/bindings/cne/msgchan.go
```diff
@@ -124,3 +135,40 @@ func (mc *MsgChannel) SendFree() int {
...
+		SendCnt:      uint64(mcInfo.send_cnt),
+		RecvCalls:    uint64(mcInfo.recv_calls),
+		RecvCnt:      uint64(mcInfo.recv_cnt),
+		RecvTimeouts: uint64(mcInfo.recv_timeouts),
+	}
+
+	return info
+}
+
+func (mc *MsgChannel) Pointers() (rx unsafe.Pointer, tx unsafe.Pointer) {
```
jeffreybshaw, line 163: Why do you need this API? Do you want to test it in the cne_test.go?

lib/usr/clib/msgchan/msgchan.h
KeithWiles: fixed

lang/go/bindings/cne/msgchan.go
KeithWiles: Added the test for this one.

lang/go/bindings/examples/tests/cne_test.go
```diff
@@ -54,6 +54,16 @@ func TestMsgChan(t *testing.T) {
 
 			mc.Close()
 		})
+		g.It("MsgChannel Info", func() {
+			mc, err := cne.NewMsgChannel(msgChanName, msgChanSize)
```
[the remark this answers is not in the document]
KeithWiles: Added the defer statements. The test if assert() failed would cleanup, but adding the defers is fine too.

jeffreybshaw (approved):

PR #130, opened 2022-07-20
fix cndpfwd with memif pmd hang

sushmasi:
cndpfwd running in client/server mode with memif pmd, after the
client instance is stopped, when server instance is tried to stop,
would hang. When the client instance exited, lport was down. But,
since the admin_state flag was not set to false, the server would
continue to tx packets and never exit.

-In pktdev_rx_burst() and pktdev_tx_burst(), packet stream state flag
was always returning true even in the case where an lport was down
and packet stream wasn't available. Now, in the case where lport is
down, 0xFFFF is returned indicating admin_state_down. Changes are
made to cndpfwd app to correctly handle the case where 0xFFFF is
returned.

-Closing of ports for all threads which was previously done during
thread_cleanup is now separated into a separate function, before
check_thread_quit.

-Handling of the admin_state_down return value in all the
tests/examples/lib that reference pktdev_rx_burst and pktdev_tx_burst
functions.

### DIFF
--- a/examples/cndpfwd/main.c
+++ b/examples/cndpfwd/main.c
@@ -32,6 +32,8 @@ struct create_txbuff_thd_priv_t {
 #define foreach_thd_lport(_t, _lp) \
     for (int _i = 0; _i < _t->lport_cnt && (_lp = _t->lports[_i]); _i++, _lp = _t->lports[_i])
 
+#define TIMEOUT_VALUE 1000 /* Number of times to wait for each usleep() time */
+
 enum thread_quit_state {
     THD_RUN = 0, /**< Thread should continue running */
     THD_QUIT,    /**< Thread should stop itself */
@@ -71,6 +73,8 @@ __tx_flush(struct fwd_port *pd, pkt_api_t api, pktmbuf_t **mbufs, uint16_t n_pkt
 {
     while (n_pkts > 0) {
         uint16_t n = __tx_burst(api, pd, mbufs, n_pkts);
+        if (n == PKTDEV_ADMIN_STATE_DOWN)
+            return n;
 
         n_pkts -= n;
         mbufs += n;
@@ -89,6 +93,8 @@ _drop_test(jcfg_lport_t *lport, struct fwd_info *fwd)
         CNE_ERR_RET("fwd_port passed in lport private data is NULL\n");
 
     n_pkts = __rx_burst(fwd->pkt_api, pd, pd->rx_mbufs, fwd->burst);
+    if (n_pkts == PKTDEV_ADMIN_STATE_DOWN)
+        return -1;
 
     if (n_pkts)
         pktmbuf_free_bulk(pd->rx_mbufs, n_pkts);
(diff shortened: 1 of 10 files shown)


examples/cndpfwd/main.c
```diff
@@ -404,11 +411,38 @@ _thread_cleanup(jcfg_info_t *j __cne_unused, void *obj, void *arg, int idx __cne
...
 
-        if (lport->umem) {
-            int i;
+static int
+_check_thread_quit(jcfg_info_t *j __cne_unused, void *obj, void *arg __cne_unused, int idx)
+{
+    jcfg_thd_t *thd = obj;
+
+    /* Make sure worker threads are done. Ignore main thread (idx=0) */
+    while (thd->quit != THD_DONE && idx > 0)
```
KeithWiles, line 424: I was looking at this forever loop and I was thinking we could add a usleep(10000); or some value to reduce the lcore utilization when it is stuck on this while loop. Plus we could do a timeout here too;

```
#define TIMEOUT_VALUE 1000 /* Number of times to wait for each usleep() time */
uint32_t timo = TIMEOUT_VALUE;

while(timo-- && thd->quit != THD_DONE && idx > 0)
    usleep(10000); /* 10ms */

if (timo < 0)
    return -1;

```
Just a thought as hanging is not good I would prefer it timeout then hang for some reason.

Need to update each place we can hang. As the code it common between examples it may exist in other examples too.

lib/cnet/eth/eth_rx.c
```diff
@@ -117,6 +117,8 @@ eth_rx_node_do(struct cne_graph *graph, struct cne_node *node, eth_rx_node_ctx_t
     /* Get pkts from port */
     nb_pkts = (node->size >= CNE_GRAPH_BURST_SIZE) ? CNE_GRAPH_BURST_SIZE : node->size;
     count   = pktdev_rx_burst(ctx->port_id, (pktmbuf_t **)node->objs, nb_pkts);
+    if (count == PKTDEV_ADMIN_STATE_DOWN)
+        return -1;
```
KeithWiles, line 121: The routine returns uint16_t so maybe we return count in this case.

lib/usr/clib/nodes/pktdev_tx.c
```diff
@@ -29,6 +29,8 @@ pktdev_tx_node_process(struct cne_graph *graph, struct cne_node *node, void **ob
 
     do {
         int cnt = pktdev_tx_burst(port, (pktmbuf_t **)objs, nb_objs);
+        if (cnt == PKTDEV_ADMIN_STATE_DOWN)
+            return -1;
```
KeithWiles, line 33: returns a uint16_t here too.

KeithWiles (changes requested):
Looks pretty good, a few minor issues.

examples/cndpfwd/main.c
sushmasi: > I was looking at this forever loop and I was thinking we could add a usleep(10000); or some value to reduce the lcore utilization when it is stuck on this while loop. Plus we could do a timeout here too;
> 
> ```
> #define TIMEOUT_VALUE 1000 /* Number of times to wait for each usleep() time */
> uint32_t timo = TIMEOUT_VALUE;
> 
> while(timo-- && thd->quit != THD_DONE && idx > 0)
>     usleep(10000); /* 10ms */
> 
> if (timo < 0)
>>>>>>>>>Isn't this always false since uint32_t?
>     return -1;
 >>>>>>>>>is this return value being used?
> ```
> 
> Just a thought as hanging is not good I would prefer it timeout then hang for some reason.
> 
> Need to update each place we can hang. As the code it common between examples it may exist in other examples too.

KeithWiles: Yes,  you are correct it should 'int timo = TIMEOUT_VALUE;' or we could move the -- to --timo and then test for zero below.

sushmasi: Done

lib/cnet/eth/eth_rx.c
sushmasi: Changed

lib/usr/clib/nodes/pktdev_tx.c
sushmasi: Changed

KeithWiles (approved):

PR #137, opened 2022-07-22
Add support for buffered TX port.

manojgop:
- Buffered TX port allows to buffer one packet at a time and transmits the
  packets once buffer is full. It can be used to forward packet arriving
  at one port to a different port by buffering the packet in destination port.

- Store ports in a map when CNE Instance is created. When application requests
  a port, return a clone (atomic reference count) of port. This also allows port
  to be safely accessed across multiple Rust threads.

### DIFF
--- a/lang/rs/apis/cne/src/port.rs
+++ b/lang/rs/apis/cne/src/port.rs
@@ -33,7 +33,7 @@ struct PortInner {
 unsafe impl Send for PortInner {}
 
 /// Port statistics.
-#[derive(Default, Debug)]
+#[derive(Clone, Default, Debug)]
 pub struct PortStats {
     /// Total number of successfully received packets.
     pub in_packets: u64,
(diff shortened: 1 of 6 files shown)


lang/rs/apis/cne/src/port.rs
```diff
@@ -262,6 +262,20 @@ impl Port {
         cne.get_port_details(port.port_index)
     }
 
+    /// Get port index of the the [port](Port).
```
maryamtahhan, line 265:
```suggestion
    /// Get port index of the [port](Port).
```

maryamtahhan (commented):
Looks good overall - just spotted a small typo

lang/rs/apis/cne/src/port.rs
manojgop: > Looks good overall - just spotted a small typo

Thanks. Updated PR

KeithWiles (approved):
LGTM, but not knowing Rust very well is a limitation of this review.

PR #141, opened 2022-07-22
fix static default builds

KeithWiles:
In the Makefile the default builds did not use the static_build flag
and would always build shared executable.


anatolyburakov:
Works with default build now, thanks!

jeffreybshaw (approved):

PR #142, opened 2022-07-22
Rust pktfwd example using CNE crate.

manojgop:
- Packet forward example using CNE crate. Supports drop/tx-only/loopback/forward mode.
  This example is similar to CNDP C pktfwd application.

- Update clap (command line parser) version in existing examples.


KeithWiles (approved):
LGTM, but I will let some other get a chance to review it before it is merged.

jeffreybshaw:
Why do we need so many different examples? Manoj, what's different between `apis/cne/examples/loopback`, `examples/echo_server`, `examples/fwd`, and `examples/pktfwd`? I think we should consolidate all these into a single example like we have for `cndpfwd`.

manojgop:
> Why do we need so many different examples? Manoj, what's different between `apis/cne/examples/loopback`, `examples/echo_server`, `examples/fwd`, and `examples/pktfwd`? I think we should consolidate all these into a single example like we have for `cndpfwd`.

1. `apis/cne/examples/loopback` is added because Rust crate needs a simple example directory within the crate to show API usage when we publish CNE crate. This is the general practice followed by many rust crates.
2. `examples/echo_server` and `examples/fwd` are elaborate examples kept at top level examples directory. Echo server compares the performance of CNE with existing popular Rust libpnet crate (Also Edwin had suggested to add this). fwd example uses CNE crate and is same as CNDP C pktfwd example. `examples/pktfwd` is old Rust example (not using CNE) and I'll remove it once this PR is merged.

jeffreybshaw:
> > Why do we need so many different examples? Manoj, what's different between `apis/cne/examples/loopback`, `examples/echo_server`, `examples/fwd`, and `examples/pktfwd`? I think we should consolidate all these into a single example like we have for `cndpfwd`.
> 
> 1. `apis/cne/examples/loopback` is added because Rust crate needs a simple example directory within the crate to show API usage when we publish CNE crate. This is the general practice followed by many rust crates.
> 2. `examples/echo_server` and `examples/fwd` are elaborate examples kept at top level examples directory. Echo server compares the performance of CNE with existing popular Rust libpnet crate (Also Edwin had suggested to add this). fwd example uses CNE crate and is same as CNDP C pktfwd example. `examples/pktfwd` is old Rust example (not using CNE) and I'll remove it once this PR is merged.

ok!

PR #144, opened 2022-07-27
add tx-only-rx mode

KeithWiles:
The originl tx-only mode would also check the RX ring and free
any packets found. The tx-only mode will now only send packets and
not check the RX ring for packets.

A new mode tx-only-rx mode has been added to do tx-only and check
the RX ring for packets and free any packets found.

### DIFF
--- a/doc/guides/sample_app_ug/cndpfwd.rst
+++ b/doc/guides/sample_app_ug/cndpfwd.rst
@@ -13,6 +13,7 @@ The cndpfwd sample application is a simple example of packet processing using CN
  * fwd
  * acl-strict
  * acl-permissive
+ * tx-only-rx (TX-Only plus draining RX ring)
 
 In forward mode ('fwd'), the destination logical port on which to forward the packet is specified by
 the last octet of the destination MAC address. It should be a valid lport number 0-N. If the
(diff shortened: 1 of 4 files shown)


jeffreybshaw:
Why do we need a new mode? What's wrong with the current "tx + rx" way? I believe we had to do it that way because of a quirk in the E810 driver that required bumping the Rx ring periodically to trigger napi thread to avoid lockup when only doing Tx.

KeithWiles:
Manoj, was asking why we were doing RX in a TX-Only routine and I explained the problem to him. In the Rust code he removed it I believe or will be removing the RX poll.

Manoj tested on E810 and I tested it on FVL NICs and did not see any lockup problem.
Do we know how to reproduce the problem?

The RX polling in tx-only mode should have only been a temp fix until we find and correct the problem, but with RX polling we are still masking the problem and we need stop masking the problem. With the new mode we can have them use it when they encounter the problem as a temp solution and possible get more information on how to fix the issue.

This is why I decided to create a new mode.

jeffreybshaw (approved):

maryamtahhan:
IIRC it was when running 64b packets at very high rates... - there still could be an issue depending on kernel version (and whatever mods have gone into the latest ice driver maybe resolve the issue).

PR #145, opened 2022-07-28
Use spin::Mutex for port instead of sync::Mutex to improve performance.

manojgop:
- Using spin mutex for port instead of sync::Mutex is improving performance
  in case of packet forward.

- Update example scripts to do incremental build.


KeithWiles (approved):
LGTM

PR #150, opened 2022-07-29
Update Rust packet forward example script.

manojgop:
…ers.

### DIFF
--- a/lang/rs/README.md
+++ b/lang/rs/README.md
@@ -1,4 +1,4 @@
-### Rust CNE Bindings, High Level API and Example Applications.
+# Rust CNE Bindings, High Level API and Example Applications
 
 Rust bindings for CNDP CNE C library.
 
@@ -47,11 +47,11 @@ Following are the steps required to build Rust bindings, high level APIs and exa
       If no command line options are passed, the script will be executed with default values used in the script.
    4. Run loopback example:  `cd apis/cne`. Execute [./run_loopback.sh](./run_loopback.sh)
    5. Run echo server example: `cd examples/echo_server`. Execute [./run_echo_server.sh](./run_echo_server.sh)
-   6. Run packet forward example: `cd examples/fwd`. Execute [./run_fwd.h](./run_fwd.h) `<drop|fwd|tx-only|lb>`
+   6. Run packet forward example: `cd examples/fwd`. Execute [./run_fwd.h](./run_fwd.h) `<JSONC config file>` `[[drop | rx-only], tx-only, [fwd | forward], [lb | loopback]]`
 
 9. To generate high level API crate documentation execute [./generate_cne_docs.sh](./generate_cne_docs.sh)
 
-### Wireguard with CNDP
+## Wireguard with CNDP
 
 Wireguard user space (implemented in Rust) uses CNDP to send and receive packets from/to user space.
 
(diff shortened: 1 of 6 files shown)


jeffreybshaw:
> …ers.
> 
> Signed-off-by: Manoj Gopalakrishnan [manoj.gopalakrishnan@intel.com](mailto:manoj.gopalakrishnan@intel.com)

Can you fix the commit message? The line length is really long, which is awkward for maintainers when reviewing the git log. Though we don't enforce this per-se, I prefer to see commit messages like the ones described in [https://www.kernel.org/doc/html/v4.10/process/submitting-patches.html#the-canonical-patch-format](the linux patch submission guide).

lang/rs/README.md
```diff
@@ -47,7 +47,7 @@ Following are the steps required to build Rust bindings, high level APIs and exa
       If no command line options are passed, the script will be executed with default values used in the script.
    4. Run loopback example:  `cd apis/cne`. Execute [./run_loopback.sh](./run_loopback.sh)
    5. Run echo server example: `cd examples/echo_server`. Execute [./run_echo_server.sh](./run_echo_server.sh)
-   6. Run packet forward example: `cd examples/fwd`. Execute [./run_fwd.h](./run_fwd.h) `<drop|fwd|tx-only|lb>`
+   6. Run packet forward example: `cd examples/fwd`. Execute [./run_fwd.h](./run_fwd.h) `<JSONC config file>` `<drop|fwd|tx-only|lb>`
```
KeithWiles, line 50: What about '[[drop | rx-only], tx-only, [fwd | forward], [lb | loopback]] we have shorter modes like fwd and lb which needs to be stated?

jeffreybshaw (commented):
Other than the commit message, this looks fine to me.

lang/rs/examples/fwd/src/main.rs
```diff
@@ -25,10 +25,10 @@ enum Mode {
     Drop,
     RxOnly,
```
KeithWiles, line 26: Strange that we do not have an alias for drop as rx-only or rx-only with alias of drop. Should we define alias here?

KeithWiles (changes requested):
A few nits.

lang/rs/examples/fwd/src/main.rs
manojgop: For RxOnly and Drop, Rust clap module automatically enum shows it as drop and rx-only. It doesn't allow to pass "Drop" or "RxOnly" in command line.

lang/rs/README.md
manojgop: I'll update the ReadME

manojgop:
> > …ers.
> > Signed-off-by: Manoj Gopalakrishnan [manoj.gopalakrishnan@intel.com](mailto:manoj.gopalakrishnan@intel.com)
> 
> Can you fix the commit message? The line length is really long, which is awkward for maintainers when reviewing the git log. Though we don't enforce this per-se, I prefer to see commit messages like the ones described in [the linux patch submission guide](https://www.kernel.org/doc/html/v4.10/process/submitting-patches.html#the-canonical-patch-format).

Updated the commit message

lang/rs/examples/fwd/src/main.rs
KeithWiles: I do not understand, but if it is not allowed then I guess it is OK.

manojgop: Rust enum variant should be in camel case.  Rust clap module will change enum variant "RXOnly" to "rx-only" by default.

KeithWiles (approved):
LGTM

PR #151, opened 2022-07-29
add a linter workflow

elzamath:
Added a GitHub action that uses the actions/super-linter GitHub
action available on the GitHub actions Marketplace
https://github.com/github/super-linter

The action will run on modified files in a pull request. Enabled
linting for Bash, C, Go, Python Rust and Markdown.

### DIFF
--- a/.github/workflows/super-linter.yml
+++ b/.github/workflows/super-linter.yml
@@ -1,0 +1,36 @@
+# This workflow executes several linters on changed files based on languages used in your code base whenever
+# you push a code or open a pull request.
+#
+# You can adjust the behavior by modifying this file.
+# For more information, see:
+# https://github.com/github/super-linter
+name: Lint Code Base
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    branches: [ "main" ]
+jobs:
+  run-lint:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+        with:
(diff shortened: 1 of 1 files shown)


jeffreybshaw:
It says the check passed, but I'm not convinced it did anything. I would have expected it to list some files in the "Files to Check:" section, right?

elzamath:
Right! It didn't do anything. That's because VALIDATE_ALL_CODEBASE: false so only edited files on this PR are checked. Since the only edited file in this PR is a yml file , the enabled linters don't run on this PR. Enabled linters are clang-format, bash, Go, Rust, Python and Markdown which don't match .yml file. Since I explicitly set the VALIDATE_<LANGUAGE> env variables to true for these languages, that means I opted to disable the rest of the linters according to the behavior mentioned here https://github.com/github/super-linter#environment-variables  I saw a linter you could enable for GitHub actions VALIDATE_GITHUB_ACTIONS: true, let me enable that and see if it runs on this PR. There also seems to be a VALIDATE_YAML env variable you can set to lint yaml files but that's probably not what we want.

jeffreybshaw:
> Right! It didn't do anything. That's because VALIDATE_ALL_CODEBASE: false so only edited files on this PR are checked. 

Ah, that makes sense. I think this is a good idea. Let's merge it and see how it goes!

jeffreybshaw (approved):

KeithWiles:
At some point we may want to make this a requirement to merging.

PR #153, opened 2022-08-13
fix pktdev_close to release internal lport data

KeithWiles:
When pktdev_close() was called the code would close the lport
but it needed to release the lport by calling pktdev_release_port().

Cleanup the a PMD probe and remove routines. In the probe routine we
now store the struct pktdev_driver pointer in the struct cne_pktdev structure
to allow for a cleaner remove or close of an lport. Remove pktdev_create_done()
routine and move the logic into the pktdev_port_setup() routine.

This would allow the Go code to be able to cleanup correctly and the PMDs
were checking if dev == NULL before calling pktdev_release_port() which is
not required.

Remove the PKTDEV_LOG() macro and use CNE_XXX macros instead.

### DIFF
--- a/lib/core/pktdev/pktdev_api.h
+++ b/lib/core/pktdev/pktdev_api.h
@@ -400,8 +400,15 @@ CNDP_API int pktdev_port_setup(lport_cfg_t *c);
  *
  * @param lport_id
  *   The lport ID to use for the removal of the lport from the system
+ * @return
+ *   -1 on error or 0 success
  */
 CNDP_API int pktdev_port_remove(int lport_id);
+
+/**
+ * Remove all lports and releases each lport resources, ignoring errors.
+ */
+CNDP_API void pktdev_port_remove_all(void);
 
 /**
  * Dump out a lport_cfg_t structure.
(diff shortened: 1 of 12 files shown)


lib/core/pktdev/pktdev_api.h
```diff
@@ -400,9 +400,16 @@ CNDP_API int pktdev_port_setup(lport_cfg_t *c);
  *
  * @param lport_id
  *   The lport ID to use for the removal of the lport from the system
+ * @return
+ *   -1 on error or 0 success
  */
 CNDP_API int pktdev_port_remove(int lport_id);
 
+/**
+ * Remove all lports and releases each lport resources, ignoreing errors.
```
jeffreybshaw, line 409: typo: ignoreing -> ignoring

lib/core/pmds/net/ring/pmd_ring.c
```diff
@@ -175,10 +183,10 @@ do_pmd_ring_create(const char *name, struct ring_internal_args *args)
     /* reserve an pktdev entry */
     dev = pktdev_allocate(name, NULL);
     if (!dev) {
-        CNE_ERR("%s: Could not pktdev_allocate\n", __FUNCTION__);
         errno = ENOSPC;
-        goto error;
+        CNE_ERR_GOTO(error, "%s: Could not pktdev_allocate\n", __FUNCTION__);
```
jeffreybshaw, line 187: __FUNCTION__ is not needed since CNE_LOG already prints it.

jeffreybshaw (approved):

PR #154, opened 2022-08-15
Go binding cleanup and example app

KeithWiles:
Here is a list on what I cleaned up:

- Simplify the usability and utilize the full JSON-c configuration.
- Remove some files that are combined into others and rework a few areas to make it easier to understand.
- Rework RxBurst/TxBurst/… to take a lportID instead of passing a method pointer.
- Cleaned up a number of allocations around LPort as this data is static after startup.
- Remove all of the underscore variables and convert them to Go variable names.
- Add the start of documentation to all of the files. Needs more work.
- Added a few more helper routines to parse a slice to packets or objects.
- Update the example/fwd application to utilize the JSON-c instead of JSON
- Remove the -port option and use the JSON-c thread section to identify the lports attached to a given thread
- Add support for using the JSON Threads section for starting threads an assigning lports to those threads.
- Use the tview Go library to display the stats in a easier to read format.
- Update the test routines to poll more than just one lport.
- Updated the cne/run_test code and script to test more functions.
- Updated the example/run_fwd script remove the -port option with other cleanups.
- More minor changes for refactoring and cleanup.


KeithWiles:
The super linter is executing on the XXX_test.go files and it should not unless we can figure out how to make it work. I do not think is a big problem, so I would like to figure out how to exclude *_test.go files.

The next problem is the Super-Linter does not appear to understand import "C" statements and that causes a number of failures.

KeithWiles:
This PR needs #153 to be complete, but I could not find how to make this PR dependent on PR 153.

elzamath:
> The super linter is executing on the XXX_test.go files and it should not unless we can figure out how to make it work. I do not think is a big problem, so I would like to figure out how to exclude *_test.go files.
> 
> The next problem is the Super-Linter does not appear to understand import "C" statements and that causes a number of failures.

setting VALIDATE_GO: false https://github.com/CloudNativeDataPlane/cndp/blob/main/.github/workflows/super-linter.yml#L33 would prevent Go files from being linted

These issues seem to be related to the failures shown by the Go linter
https://github.com/golangci/golangci-lint/issues/1176 related to the 
```
could not import C (cgo preprocessing failed) (typecheck)
import "C"
```

https://github.com/github/super-linter/issues/1599 - `undeclared name: `System` (typecheck)`
 related to the the undeclared name failure

lang/go/bindings/cne/lport.go
```diff
@@ -0,0 +1,304 @@
...
+type LPortStats struct {
+	InPackets     uint64 // Number of input packets
+	InBytes       uint64 // Number of input bytes
+	InErrors      uint64 // Number of input errors
+	InMissed      uint64 // Number of missed input packets
+	RxInvalid     uint64 // Number of invalid Rx packets
+	OutPackets    uint64 // Number of output packets
+	OutBytes      uint64 // Number of output bytes
+	OutErrors     uint64 // Number of output errors
+	OutDropped    uint64 // Number of dropper output packets
```
maryamtahhan, line 50:
```suggestion
	OutDropped    uint64 // Number of dropped output packets
```

lang/go/bindings/cne/lport.go
```diff
@@ -0,0 +1,304 @@
...
+	pcfg.bufsz = C.uint(lportUmem.BufSize)
+	pcfg.rx_nb_desc = C.uint(lportUmem.RxDesc)
+	pcfg.tx_nb_desc = C.uint(lportUmem.TxDesc)
+
+	lportNameArr := strings.Split(lportName, ":")
+	netdevName := lportNameArr[0]
+
+	for i := 0; i < len(netdevName) && i < C.LPORT_NAME_LEN; i++ {
+		pcfg.ifname[i] = C.char(netdevName[i])
+		pcfg.name[i] = C.char(netdevName[i])
```
maryamtahhan, line 92: I think `pcfg.name` would be set to `lportName` if we follow the c example?

lang/go/bindings/cne/lport.go
```diff
@@ -0,0 +1,304 @@
...
+
+	return lportList
+}
+
+// LPortByName returns the LPort for the given name
+func (sys *System) LPortByName(name string) *LPort {
+
+	lport, ok := lportMap[name]
+	if !ok {
+		fmt.Printf("LPort %#v not found\n", name)
```
maryamtahhan, line 205: Should we use `log.Printf` rather than `fmt.Printf`?

lang/go/bindings/cne/lport.go
```diff
@@ -0,0 +1,304 @@
...
+	return lports
+}
+
+// Name returns the name of the given LPort
+func (p *LPort) Name() string {
+
+	return p.name
+}
+
+// Name returns the name of the given LPort
```
maryamtahhan, line 232:
```suggestion
// NetdevName returns the netdev name of the given LPort
```

lang/go/bindings/cne/lport.go
```diff
@@ -0,0 +1,304 @@
...
+	return lport
+}
+
+// LPortsByName return a list of LPort structure pointers
+func (sys *System) LPortsByName(names []string) []*LPort {
+
+	var lports []*LPort
+
+	for _, name := range names {
+		lport := sys.LPortByName(name)
```
maryamtahhan, line 217: just a style nit... in some places we do an initialisation like here then follow up with the nil check... then in other places we do the one liner variation like:

```
	if ret := C.pktdev_stats_get(p.lportId, &stats); ret < 0 {
		return nil, fmt.Errorf("LPortStats failed with error code %d", ret)
	}
```

Maybe it would be nice to pick a style... it's not a show stopper/blocker - just a nit

lang/go/bindings/cne/lport.go
```diff
@@ -0,0 +1,304 @@
...
+	p.prevOutPackets = ps.OutPackets
+
+	return ps, nil
+}
+
+// ResetLPortStats resets the LPort statistics
+func (p *LPort) ResetLPortStats() error {
+
+	if ret := C.pktdev_stats_reset(p.lportId); ret < 0 {
+		return fmt.Errorf("ResetLPortStats failed with error code %d", ret)
```
maryamtahhan, line 278: just again, I'm wondering about log vs fmt...

maryamtahhan (commented):
Just a start.... 
got as far as lport_test.go (will keep tipping away at it)

lang/go/bindings/cne/jcfg.go
```diff
@@ -0,0 +1,373 @@
...
+}
+
+// LPortInfo is the JSON LPort information data structure(s)
+type LPortInfo struct {
+	Pmd             string `json:"pmd"`               // PMD name
+	Qid             uint16 `json:"qid"`               // QID number
+	Umem            string `json:"umem"`              // UMEM name string for this LPort to use
+	Region          int    `json:"region"`            // The UMEM region index value
+	Description     string `json:"description"`       // Description of the LPort
+	BusyPolling     bool   `json:"busy_polling"`      // "busy_poll" tag name is not supported
```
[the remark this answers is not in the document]
KeithWiles: We do need to support busy timeout/budget. I will look into adding it, as the previous code did not attempt to use that values. The BusyPolling the code does set the flag and I need to verify is works.

lang/go/bindings/cne/lport.go
KeithWiles: This one we are returning a error object so must be a fmt.Errorf

maryamtahhan: d'oh - of course

lang/go/bindings/cne/jcfg.go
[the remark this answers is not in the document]
KeithWiles: Looking at the code we do use the Busy* fields in the round line 100, is this what you wanted?

lang/go/bindings/cne/lport.go
KeithWiles: Let me look at them, I did try to fix up the ones I found to be the same when able to.

lang/go/bindings/cne/jcfg.go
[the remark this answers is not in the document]
maryamtahhan: no - It's fine. I see we do check and use it in the code below as you mention so is the `is not supported` part needed in the comment?

lang/go/bindings/cne/system.go
```diff
@@ -9,55 +9,150 @@ package cne
...
+}
+
+// Close the binding layer and cleanup UMEM and LPorts
+func (sys *System) Close() error {
+
+	defer sys.UnregisterThread(sys.tid)
+
+	if err := sys.cleanupUMEM(); err != nil {
+		return fmt.Errorf("error cleaning up mempool %w", err)
+	}
```
maryamtahhan, line 139: do we need a blank like before the next if statement?

lang/go/bindings/cne/system_test.go
```diff
@@ -5,15 +5,83 @@
...
+
+			// Read the JSON-C file into a single byte slice for later parsing
+			if bytes, err = io.ReadAll(configFile); err != nil {
+				t.Errorf("unable to load config file: %#v: %v", configStr, err)
+			}
+
+			// Open with a JSON-C string
+			if cneSys, err := OpenWithString(string(bytes)); err != nil {
+				t.Errorf("error parsing JSON string: %v", err)
+			} else {
```
maryamtahhan, line 35: do we need the else here?

lang/go/bindings/cne/system_test.go
```diff
@@ -5,15 +5,83 @@
...
+				time.Sleep(time.Second)
+			}
+		}
+	})
+
+	t.Run("OpenWithFile", func(t *testing.T) {
+		// Open with a JSON-C filename
+		if cneSys, err := OpenWithFile(*configStr); err != nil {
+			t.Errorf("error parsing JSON string %#v: %v", configStr, err)
+		} else {
```
maryamtahhan, line 46: and here?

lang/go/bindings/cne/umem.go
```diff
@@ -0,0 +1,154 @@
...
+		ri.bufCnt = umem.Regions[i]
+		ri.addr = (*C.char)(umemAddr)
+		umemAddr = (unsafe.Pointer)((uintptr(umemAddr) + uintptr(ri.bufCnt*umem.BufSize)))
+
+		bufcnt := C.uint(ri.bufCnt)
+		bufsz := C.uint(umem.BufSize)
+		cache := C.uint(sys.jcfg.Defaults().Cache)
+
+		pi := (*C.pktmbuf_info_t)(C.pktmbuf_pool_create(ri.addr, bufcnt, bufsz, cache, nil))
+		if pi == nil {
```
maryamtahhan, line 77: if we fail here should we free mmapPtr before returning the error or do we just want to say that we failed for a region in the umem?

maryamtahhan (commented):
got as far as the examples dir - will try to close off the rest tomorrow

lang/go/bindings/cne/jcfg.go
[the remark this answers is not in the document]
KeithWiles: removed the comment

lang/go/bindings/cne/lring.go
```diff
@@ -125,7 +124,7 @@ func LRingCreateElem(name string, elementSz, count uint, strFlags []string) (*Lo
 	return lr, nil
 }
```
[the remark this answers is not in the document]
KeithWiles: I saw only one missing line around line 107

lang/go/bindings/cne/run_test
```diff
@@ -1,21 +1,72 @@
 #!/bin/bash
 
-project_path="${PROJECT_PATH:-/root/git/networking.dataplane.cndp.cndp}"
+cdir=$(pwd)
+PROJECT_PATH="${cdir}/../../../.."
 
-go build
+precmd="sudo -E LD_LIBRARY_PATH=$PROJECT_PATH/usr/local/lib/x86_64-linux-gnu LD_PRELOAD=$PROJECT_PATH/usr/local/lib/x86_64-linux-gnu/libpmd_af_xdp.so"
+cmdstring="$precmd ./cne.test -test.v -config ../examples/fwd/fwd.jsonc"
+#cmdcover="$precmd ./cne.test -test.v -test.coverprofile cne.profile -config ../examples/fwd/fwd.jsonc"
```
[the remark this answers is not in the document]
KeithWiles: I have not finished the different types of tests, so I can remove it.

lang/go/bindings/cne/system.go
```diff
@@ -9,55 +9,150 @@ package cne
...
+	sys := &System{}
+
+	sys.tid = sys.RegisterThread("main")
+
+	switch v := cfg.(type) {
+	case *Config: //  Initialize the binding layer with a cne.Config structure
+		jcfg := v
+
+		if err := jcfg.validateConfig(); err != nil {
+			sys.UnregisterThread(sys.tid)
```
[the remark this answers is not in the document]
KeithWiles: The error object is returned and that should be printed.

lang/go/bindings/cne/system_test.go
```diff
@@ -5,15 +5,83 @@
...
+
+			if jcfg, err := processConfig(jsonc.ToJSON(bytes)); err != nil {
+				t.Errorf("unable to process config file: %#v: %v", configStr, err)
+			} else {
+				jcfg.valuesToUnitMultipliers()
+
+				// Open with a cne.Config structure
+				if cneSys, err := OpenWithConfig(jcfg); err != nil {
+					t.Errorf("error parsing JSON string %#v: %v", configStr, err)
+				} else {
```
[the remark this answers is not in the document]
KeithWiles: The cneSys variable goes out of scope because it is created within the 'if' construct, is that not true?

lang/go/bindings/cne/system_test.go
KeithWiles: Same as above comment.

lang/go/bindings/cne/system_test.go
KeithWiles: Same as above

lang/go/bindings/cne/lring.go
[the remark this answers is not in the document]
maryamtahhan: that was the missing line, it wouldn't let me add the comment at that point

lang/go/bindings/cne/system_test.go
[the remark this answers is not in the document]
maryamtahhan: yes it does

KeithWiles:
Are we all done reviewing this one?
@maryamtahhan @jeffreybshaw @sushmasi

jeffreybshaw:
> 

If @maryamtahhan approves, let's merge this.

maryamtahhan (approved):
sorry for the delay. LGTM

PR #156, opened 2022-08-16
remove pktdev_port_remove calls in the Rust code

KeithWiles:
Remove the calls to pktdev_port_remove() and fixup the patch
files using editdiff tool in Linux. The tools seems to remove
the index lines, which are Git index values. I do not believe
these are required for patching wireguard code but are only used
for referening Git IDs.

### DIFF
--- a/lang/rs/apis/cne/src/config.rs
+++ b/lang/rs/apis/cne/src/config.rs
@@ -17,8 +17,8 @@ use std::result::Result;
 
 use cne_sys::bindings::{
     _pktmbuf_info_name_set, lport_cfg_t, mmap_addr, mmap_alloc, mmap_free, mmap_size, mmap_t,
-    mmap_type_by_name, pktdev_close, pktdev_port_remove, pktdev_port_setup, pktmbuf_destroy,
-    pktmbuf_info_t, pktmbuf_pool_create, uds_info_t, udsc_close, udsc_handshake, xskdev_info_t,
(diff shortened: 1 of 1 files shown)


lang/rs/wireguard/patch/0001-Integarte-CNDP-Cloud-Native-Data-Plane-with-Wireguar.patch
```diff
@@ -99,7 +99,6 @@ index 802feda..7ca64af 100644
 +**/bindings.rs
```
manojgop, line 99: I don't think we need to update the wireguard patches. The new wireguard patch uses Rust CNE library . So `pktdev_port_remove()` won't be called from wireguard

manojgop:
Existing wireguard patches need not be updated in this commit. Older wireguard patches which calls `pktdev_port_remove()` is overridden by new patch which uses Rust CNE library. So removing the deprecated function `pktdev_port_remove()`  from Rust CNE library alone should suffice.

lang/rs/wireguard/patch/0001-Integarte-CNDP-Cloud-Native-Data-Plane-with-Wireguar.patch
KeithWiles: Manoj stated we do not need to update the patch file, so reverted the changes.

jeffreybshaw (approved):

PR #159, opened 2022-08-17
fix tx-only packet format

KeithWiles:
The packet format used in tx-only mode was missing the ethertype. Replaced
with a IPv4/UDP frame.

The Go version will be fixed in PR #154 

### DIFF
--- a/examples/cndpfwd/main.c
+++ b/examples/cndpfwd/main.c
@@ -194,14 +194,27 @@ _txonly_test(jcfg_lport_t *lport, struct fwd_info *fwd)
             pktmbuf_t *xb = tx_mbufs[j];
             uint64_t *p   = pktmbuf_mtod(xb, uint64_t *);
 
-            p[0]                 = 0xfd3c78299efefd3c;
-            p[1]                 = 0x00450008b82c9efe;
-            p[2]                 = 0x110400004f122e00;
-            p[3]                 = 0xa8c00100a8c01e22;
-            p[4]                 = 0x1a002e16d2040101;
-            p[5]                 = 0x706f6e6d6c6b9a9e;
-            p[6]                 = 0x7877767574737271;
-            p[7]                 = 0x31307a79;
+            /*
+             * IPv4/UDP 64 byte packet
+             * Port Src/Dest       :           1234/ 5678
+             * Pkt Type            :           IPv4 / UDP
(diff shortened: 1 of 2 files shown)


jeffreybshaw (dismissed):

examples/cndpfwd/main.c
```diff
@@ -194,14 +194,29 @@ _txonly_test(jcfg_lport_t *lport, struct fwd_info *fwd)
...
-            p[3]                 = 0xa8c00100a8c01e22;
-            p[4]                 = 0x1a002e16d2040101;
-            p[5]                 = 0x706f6e6d6c6b9a9e;
-            p[6]                 = 0x7877767574737271;
-            p[7]                 = 0x31307a79;
+            /*
+             * IPv4/UDP 64 byte packet
+             * TTL/Port Src/Dest   :       64/ 1234/ 5678
+             * Pkt Type:VLAN ID    :      IPv4 / UDP:0001
+             * 802.1p CoS/DSCP/IPP :            0/  0/  0
```
jeffreybshaw, line 201: Minor nit: why do you mention VLAN and VxLAN in the comments here? This is just a normal IP/UDP packet, right?

lang/rs/examples/fwd/src/main.rs
```diff
@@ -278,7 +278,21 @@ fn tx_only(port: &Port, pkts: &mut [Packet]) -> Result<(), CneError> {
     let alloc_pkts = port.prepare_tx_packets(pkts)? as usize;
 
     // Fill a byte array with some values.
-    let input = "fd3c78299efefd3c00450008b82c9efe110400004f122e00a8c00100a8c01e221a002e16d2040101706f6e6d6c6b9a9e787776757473727131307a79";
+    // IPv4/UDP 64 byte packet
+    // TTL/Port Src/Dest   :       64/ 1234/ 5678
```
jeffreybshaw, line 282: same comment as in main.c, this is just a regular packet, but you're talking about VLAN and VxLAN in the comment which is a little confusing.

examples/cndpfwd/main.c
KeithWiles: I created this packet with pktgen and then I just grabbed the screen text for the packet. I can remove it if you like.

jeffreybshaw (approved):

PR #160, opened 2022-08-17
simulate xdp port stats reset

KeithWiles:
When the xskdev_stats_reset() is called the lport_stats_t is set to zero,
but the XDP stats are not reset as they are handled by the kernel. This
means the counter we display like imissed or inpacket do not appear to
be reset to zero. We can not reset the XDP stats so keep a copy on init
of the port and after a reset of the stats on a lport.

Add some logic to simulate the XDP stats are reset by keeping a original
set of stats from the XDP counters and substract the difference to
get the correct value after a reset call.

### DIFF
--- a/lib/core/xskdev/xskdev.c
+++ b/lib/core/xskdev/xskdev.c
@@ -1049,20 +1049,36 @@ xskdev_stats_get(xskdev_info_t *xi, lport_stats_t *stats)
         CNE_ERR_RET("getsockopt() failed for XDP_STATISTICS for fd %d, %s, %d, %d.\n", fd,
                     xi->ifname, xi->prog_id, xi->if_index);
 
-    stats->imissed += xdp_stats.rx_dropped;
-    stats->rx_invalid += xdp_stats.rx_invalid_descs;
-    stats->tx_invalid += xdp_stats.tx_invalid_descs;
-
+    /* Adjust statistics to be consistent after a xskdev_stats_reset() call */
+    stats->imissed    = (xdp_stats.rx_dropped - xi->orig_stats.rx_dropped);
+    stats->rx_invalid = (xdp_stats.rx_invalid_descs - xi->orig_stats.rx_invalid_descs);
+    stats->tx_invalid = (xdp_stats.tx_invalid_descs - xi->orig_stats.tx_invalid_descs);
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(5, 9, 0)
+    stats->rx_ring_full = (xdp_stats.rx_ring_full - xi->orig_stats.rx_ring_full);
+    stats->rx_fill_ring_empty =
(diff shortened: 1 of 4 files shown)


lib/core/xskdev/xskdev.c
```diff
@@ -974,6 +974,17 @@ xskdev_socket_create(struct lport_cfg *c)
     if (fq_reserved(xi, umem->fq_size) < 0)
         CNE_ERR_GOTO(err, "Failed reserved fill of FQ\n");
 
+    do {
```
jeffreybshaw, line 977: What's with the "do {} while (0)", just to create a separate block for your local variables? Can you define your local variables at the top of the function or create a new function to do this?

lib/core/xskdev/xskdev.c
```diff
@@ -974,6 +974,17 @@ xskdev_socket_create(struct lport_cfg *c)
...
+        socklen_t optlen = sizeof(struct xdp_statistics);
+        int ret, fd;
+
+        fd  = xsk_socket__fd(xi->rxq.xsk);
+        ret = getsockopt(fd, SOL_XDP, XDP_STATISTICS, &xi->orig_stats, &optlen);
+        if (ret != 0)
+            CNE_NULL_RET("getsockopt() failed for XDP_STATISTICS for fd %d, %s, %d, %d.\n", fd,
```
jeffreybshaw, line 984: goto err

lib/core/xskdev/xskdev.c
```diff
@@ -974,6 +974,17 @@ xskdev_socket_create(struct lport_cfg *c)
     if (fq_reserved(xi, umem->fq_size) < 0)
         CNE_ERR_GOTO(err, "Failed reserved fill of FQ\n");
 
+    do {
+        socklen_t optlen = sizeof(struct xdp_statistics);
+        int ret, fd;
+
+        fd  = xsk_socket__fd(xi->rxq.xsk);
+        ret = getsockopt(fd, SOL_XDP, XDP_STATISTICS, &xi->orig_stats, &optlen);
```
jeffreybshaw, line 982: Are you sure we need to get the stats when we create a socket? I would think any new socket would already have zeroed  stats, right?

KeithWiles: Most likely, I was thinking about how I did it in pktgen and in that case I did need to grab the stats from the NIC.

I will remove this one and assume the stats are zero.

KeithWiles: Removed the code and assume the XDP stats are zero.

lib/core/xskdev/xskdev.c
KeithWiles: This code was removed.

lib/core/xskdev/xskdev.c
KeithWiles: removed

jeffreybshaw (approved):

PR #163, opened 2022-08-18
idle manager to enable interrupt mode polling

KeithWiles:
The current design in CNDP is 100% polling of the receive ring in AF_XDP
or PMD which means burning core power. The patch adds two parameters to
the jsonc file to control polling idle_timeout and intr_timeout
in thread section.

The idlemgr library handles managing epoll and calling epoll_wait when the
idle_timeout value has been met, then it calls epoll_wait(). The idlemgr
operates on threads, using file descriptors as the method to wakeup a sleeping
thread. The caller needs to add file descriptors to the idlemgr instance and
then call idlemgr_process() with a flag if the thread thinks it is idle or
not.

The first one gives the number of milliseconds to wait for the RX ring to
be idle, then call epoll() with a timeout of intr_timeout. When no Rx traffic
for a given time the idlemgr will call epoll(), which reduces the lcore load to
effectively zero and only waking up when packets arrive or a timeout occurs.

In testing of performance it appears to be very little impact when interrupt
mode is enabled compared to when it is not enabled. Added some counters to help
determine how the new mode is operating.

The idle_timeout value in the jsonc file for given thread is how this feature
is controlled. If not defined or set to zero interrupt mode is disabled. When
set to a non zero value will enable interrupt mode. The intr_timeout value
is only used if idle_timeout is non-zero and will be used in the poll() call
as the timeout value. Each of these values are in milliseconds.


lib/core/xskdev/xskdev.c
```diff
@@ -363,9 +371,20 @@ xskdev_rx_burst_default(void *_xi, void **bufs, uint16_t nb_pkts)
             xi->stats.rx_poll_wakeup++;
             (void)poll(&rxq->fds, 1, POLL_TIMEOUT);
         }
+
+        if (xi->idle_timestamp && cne_rdtsc_precise() > xi->idle_timestamp) {
+            if (poll(&rxq->fds, 1, xi->intr_timeout) == 0)
```
jeffreybshaw, line 376: Can you test this with a single thread handling more than one XSK? Make sure there is *no* traffic on all XSK except for one, and verify that the socket wakes up? I'm suspect this won't work when you have traffic on one socket, but not another, because you'll get stuck polling on the idle socket.

This was one of the reasons we changed the original POLL_TIMEOUT from 1000 to 0.

I think we should add all the FDs for all the AF_XDP sockets handled by a particular thread to an epoll() group, and then wait to get woken up. This won't work if the thread needs to handle non AF_XDP sockets (so if its using pktdev we need to make sure only pmd_af_xdp is used).

lib/usr/app/jcfg/jcfg_lport_group.c
```diff
@@ -247,6 +247,8 @@ setup_lport(jcfg_data_t *data, jcfg_lport_group_t *lpg, const char *netdev, uint
     lport->umem         = lpg->umem;
     lport->busy_timeout = lpg->busy_timeout;
     lport->busy_budget  = lpg->busy_budget;
+    lport->idle_timeout = lpg->idle_timeout;
+    lport->intr_timeout = lpg->intr_timeout;
```
jeffreybshaw, line 251: These values need to be parsed in https://github.com/CloudNativeDataPlane/cndp/blob/main/lib/usr/app/jcfg/jcfg_lport_group.c#L703

lib/core/xskdev/xskdev.c
KeithWiles: A single thread polling more then one port will hang as you stated. The xskdev driver does not know about the other lports being polled by the same thread.

I believe this means the xskdev driver cannot be the point we do the poll() call and it needs to be at the thread level not the lport level. I am thinking this means we create library or functions to handle calling epoll() at the application level.

I know the developer could do this himself, but I think a set of APIs would make it easier to manage. This means the intr_timeout and idle_timeout values needs to be removed from jsonc lports and lport_groups an into the thread section.

We then create an epoll() adding each of the af_xdp socket fds and any other fd the thread wants to wait on. Need a bit more thought about this option. May need to figure out how to get the fd values for each lport.

KeithWiles:
This needs to be change to be thread based and not lport based, so converting this to draft status for now.

I have a new version based on threads and lport fd as the triggers, but it is a bigger task of work.

KeithWiles:
Will break this PR down when we get closer to a good solution.

lib/usr/app/jcfg/jcfg_lport_group.c
KeithWiles: not required anymore

lib/core/xskdev/xskdev.c
KeithWiles: reworked how interrupts and file descriptors are handled.

KeithWiles:
This PR is being closed and replaced by the new PR #166

nickcreate2021:
> The current design in CNDP is 100% polling of the receive ring in AF_XDP or PMD which means burning core power. The patch adds two parameters to the jsonc file to control polling idle_timeout and intr_timeout in thread section.
> 
> The idlemgr library handles managing epoll and calling epoll_wait when the idle_timeout value has been met, then it calls epoll_wait(). The idlemgr operates on threads, using file descriptors as the method to wakeup a sleeping thread. The caller needs to add file descriptors to the idlemgr instance and then call idlemgr_process() with a flag if the thread thinks it is idle or not.
> 
> The first one gives the number of milliseconds to wait for the RX ring to be idle, then call epoll() with a timeout of intr_timeout. When no Rx traffic for a given time the idlemgr will call epoll(), which reduces the lcore load to effectively zero and only waking up when packets arrive or a timeout occurs.
> 
> In testing of performance it appears to be very little impact when interrupt mode is enabled compared to when it is not enabled. Added some counters to help determine how the new mode is operating.
> 
> The idle_timeout value in the jsonc file for given thread is how this feature is controlled. If not defined or set to zero interrupt mode is disabled. When set to a non zero value will enable interrupt mode. The intr_timeout value is only used if idle_timeout is non-zero and will be used in the poll() call as the timeout value. Each of these values are in milliseconds. Signed-off-by: Keith Wiles [keith.wiles@intel.com](mailto:keith.wiles@intel.com)

Does this function has the requirement of the kernel version?when  I run it under 4.18 kernel ,epoll_ctl(imgr->epoll_fd, EPOLL_CTL_ADD, fd, &event) systemcall will let the OS corruption and restart.

KeithWiles:
I do not believe a specific kernel version is requirement. When you disable idle manager I assume the restart does not occur?

nickcreate2021:
> Contributor

YES,When I disable idle manager ,the restart does not occur。

KeithWiles:
In the idlemgr_add() routine you can pass zero to this function for eflags and event.events == 0 then it is replaced with EPOLLIN the default value.

PR #167, opened 2022-08-23
Fix go bindings lport setup

nwaples:
The full lport name was not being copied to the lport_cfg_t.
It was being trunctated to the length of the network device name.

Signed-off-by: Nicholas Waples <nwaples@gmail.com>


KeithWiles (approved):
thanks, LGTM

PR #169, opened 2022-08-25
fix go binding code to validate json-c file

KeithWiles:
When the OpenWithFile() is called with an invalid config file path/filename
it would cause an exception. Now the error is returned to the caller to
process correctly.

### DIFF
--- a/lang/go/bindings/cne/system.go
+++ b/lang/go/bindings/cne/system.go
@@ -116,11 +116,15 @@ func OpenWithFile(path string) (*System, error) {
 	if strings.HasPrefix(str, "{") {
 		err = fmt.Errorf("path does not appear to be a valid filepath")
 	} else {
-		if configFile, err := os.Open(str); err == nil {
+		var configFile *os.File
+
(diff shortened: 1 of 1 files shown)


lang/go/bindings/cne/system.go
```diff
@@ -123,6 +123,8 @@ func OpenWithFile(path string) (*System, error) {
 			if bytes, err := io.ReadAll(configFile); err == nil {
 				return open(jsonc.ToJSON(bytes))
 			}
+		} else {
+			return nil, err
```
jeffreybshaw, line 127: Wouldn't this just fall through to teh end of the function which returns nil, err?

KeithWiles: It appears the Go scoping rules have bit me. I assumed the declare of err at top of the function, that the os.Open() line would not create a new instance, but after testing it seems it does :-( Maybe something I am not understanding.

I reworked the code a bit to remove all of the := and that also seems to fix the problem.

jeffreybshaw (approved):

PR #176, opened 2022-08-29
Feat txgen latency

maryamtahhan:
Leaving as a draft request till we've discussed the pktmbuf offset issue I've encountered

### DIFF
--- a/usrtools/txgen/app/cli-functions.c
+++ b/usrtools/txgen/app/cli-functions.c
@@ -1,5 +1,6 @@
 /* SPDX-License-Identifier: BSD-3-Clause
- * Copyright (c) 2022 Intel Corporation.
+ * Copyright (c) 2019-2022 Intel Corporation
+ * Copyright (c) 2022 Red Hat, Inc.
  */
 
 #include "cli-functions.h"
@@ -72,6 +73,63 @@ static const char *status_help[] = {
     "dport|"        /*  5 */ \
     "ttl"           /*  6 */
 
+
+/**
+ *
+ * single_set_latsamp_params - Set the port latency sampler parameters
+ *
+ * DESCRIPTION
+ * Set the given port list with the given latency sampler parameters
+ *
+ * RETURNS: N/A
+ *
+ * SEE ALSO:
+ */
+
+void
+single_set_latsampler_params(port_info_t *info, char *type, uint32_t num_samples,
+                             uint32_t sampling_rate, char outfile[])
+{
+    FILE *fp = NULL;
+    uint32_t sampler_type;
+
+    /* Stop if latency sampler is running */
+    if (txgen_tst_port_flags(info, SAMPLING_LATENCIES)) {
+        CNE_WARN("Latency sampler is already running, stop it first!");
+        return;
+    }
+    /* Validate sampler type*/
+    if (!strcasecmp(type, "simple"))
+        sampler_type = LATSAMPLER_SIMPLE;
+    else if (!strcasecmp(type, "poisson"))
+        sampler_type = LATSAMPLER_POISSON;
+    else {
+        CNE_ERR("Unknown latsampler type %s! Valid values: simple, poisson", type);
+        return;
+    }
+
+    /* Validate file path */
+    fp = fopen(outfile, "w+");
+    if (fp == NULL) {
+        CNE_ERR("Cannot write to file path %s!", outfile);
+        return;
+    }
+    fclose(fp);
+
+    if (num_samples > MAX_LATENCY_ENTRIES) {
+        CNE_ERR("Too many samples requested. Max %d!", MAX_LATENCY_ENTRIES);
+        return;
+    }
+
+    info->latsamp_type        = sampler_type;
+    info->latsamp_rate        = sampling_rate;
+    info->latsamp_num_samples = num_samples;
+    strcpy(info->latsamp_outfile, outfile);
+
+    txgen_packet_ctor(info);
+}
+
+
 static struct cli_map set_map[] = {
     {10, "set %P %|" set_types " %d"},
     {20, "set %P type ipv4"},
@@ -82,6 +140,7 @@ static struct cli_map set_map[] = {
     {25, "set %P user pattern %s"},
     {30, "set %P src ip %4"},
     {31, "set %P dst ip %4"},
+    {100, "set %P latsampler %|simple|poisson %d %d %s"},
     {-1, NULL}
     };
 // clang-format on
@@ -111,6 +170,11 @@ static const char *set_help[] = {
     "set <portlist> user pattern <string> - A 16 byte string, must set 'pattern user' command",
     "set <portlist> [src|dst] ip ipaddr - Set IP addresses, Source must include network mask "
     "e.g. 10.1.2.3/24",
+    "set <portlist> latsampler [simple|poisson] <num-samples> <rate> <outfile>    - Set latency "
+    "sampler parameters",
+    "        num-samples: number of samples.",
+    "        rate: sampling rate i.e., samples per second.",
+    "        outfile: path to output file to dump all sampled latencies",
     CLI_HELP_PAUSE,
     NULL};
 
@@ -123,6 +187,7 @@ set_cmd(int argc, char **argv)
     struct cli_map *m;
     struct in_addr ip;
     int prefixlen;
+    uint32_t u1, u2;
 
     m = cli_mapping(set_map, argc, argv);
     if (!m)
@@ -210,6 +275,12 @@ set_cmd(int argc, char **argv)
             foreach_port(portlist,
                 single_set_ipaddr(info, 'd', &ip, 0));
             break;
+        case 100:
+            u1 = strtol(argv[4], NULL, 0);
+            u2 = strtol(argv[5], NULL, 0);
+            foreach_port(portlist,
+                single_set_latsampler_params(info, argv[3], u1, u2, argv[6]));
+            break;
         default:
             return cli_cmd_error("Command invalid", "Set", argc, argv);
     }
@@ -276,7 +347,11 @@ pcap_cmd(int argc, char **argv)
     return 0;
 }
 
-static struct cli_map start_map[] = {{10, "start %P"}, {20, "stop %P"}, {-1, NULL}};
+static struct cli_map start_map[] = {{10, "start %P"},
+                                     {20, "stop %P"},
+                                     {50, "start %P latsampler"},
+                                     {60, "stop %P latsampler"},
+                                     {-1, NULL}};
 
 static const char *start_help[] = {
     "",
@@ -284,6 +359,9 @@ static const char *start_help[] = {
     "stop <portlist>                    - Stop transmitting packets",
     "stp                                - Stop all lports from transmitting",
     "str                                - Start all lports transmitting",
+    "start <portlist> latsampler        - Start latency sampler, make sure to set sampling "
+    "parameters before starting",
+    "stop <portlist> latsampler            - Stop latency sampler, dumps to file if specified",
     CLI_HELP_PAUSE,
     NULL};
 
@@ -305,6 +383,12 @@ start_stop_cmd(int argc, char **argv)
         break;
     case 20:
         foreach_port(portlist, txgen_stop_transmitting(info));
+        break;
+    case 50:
+        foreach_port(portlist, txgen_start_latency_sampler(info));
+        break;
+    case 60:
+        foreach_port(portlist, txgen_stop_latency_sampler(info));
         break;
     default:
         return cli_cmd_error("Start/Stop command invalid", "Start", argc, argv);
@@ -504,6 +588,40 @@ misc_cmd(int argc, char **argv)
     return 0;
 }
 
+static struct cli_map page_map[] = {{10, "page %d"},
+                                    {11, "page "
+                                         "%|main|pcap|latency"},
+                                    {-1, NULL}};
+
+static const char *page_help[] = {
+    "",
+    "page [0-2]                         - Show the port pages or configuration or sequence page",
+    "page main                          - Display page zero",
+    "page pcap                          - Display the pcap page",
+    "page latency                       - Display the latency page",
+    CLI_HELP_PAUSE,
+    NULL};
+
+static int
+page_cmd(int argc, char **argv)
+{
+    struct cli_map *m;
+
+    m = cli_mapping(page_map, argc, argv);
+    if (!m)
+        return cli_cmd_error("Page invalid command", "Page", argc, argv);
+
+    switch (m->index) {
+    case 10:
+    case 11:
+        txgen_set_page(argv[1]);
+        break;
+    default:
+        return cli_cmd_error("Page invalid command", "Page", argc, argv);
+    }
+    return 0;
+}
+
 /**********************************************************/
 /**********************************************************/
 /****** CONTEXT (list of instruction) */
@@ -533,6 +651,7 @@ static struct cli_tree default_tree[] = {
     c_alias("str", "start all", "start all lports sending packets"),
     c_alias("stp", "stop all", "stop all lports sending packets"),
     c_cmd("pcap", pcap_cmd, "pcap commands"),
+    c_cmd("page", page_cmd, "change page displays"),
     c_cmd("set", set_cmd, "set a number of options"),
 
     c_alias("on", "enable screen", "Enable screen updates"),
@@ -552,6 +671,7 @@ init_tree(void)
         return -1;
 
     cli_help_add("Title", NULL, title_help);
+    cli_help_add("Page", page_map, page_help);
     cli_help_add("Enable", enable_map, enable_help);
     cli_help_add("Set", set_map, set_help);
     cli_help_add("PCAP", pcap_map, pcap_help);
--- a/usrtools/txgen/app/cli-functions.c
+++ b/usrtools/txgen/app/cli-functions.c
@@ -353,6 +353,7 @@ static struct cli_map start_map[] = {{10, "start %P"},
                                      {60, "stop %P  %|latsampler|lat|latency"},
                                      {-1, NULL}};
 
+// clang-format off
 static const char *start_help[] = {
     "",
     "start <portlist>                         - Start transmitting packets",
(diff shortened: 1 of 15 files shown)


usrtools/txgen/app/port-cfg.h
```diff
@@ -103,11 +120,31 @@ typedef struct port_info_s {
     fill_t fill_pattern_type;             /**< Type of pattern to fill with */
     char user_pattern[USER_PATTERN_SIZE]; /**< User set pattern values */
 
+    uint32_t latency_nb_pkts;
```
maryamtahhan, line 123: Should add doxygen descriptions

usrtools/txgen/app/txgen.c
```diff
@@ -136,6 +145,77 @@ txgen_fill_pattern(uint8_t *p, uint32_t len, uint32_t type, char *user)
...
+    return tstamp;
+}
+
+static __inline__ tstamp_t *
+txgen_tstamp_pointer_rx(port_info_t *info, pktmbuf_t *m)
+{
+    tstamp_t *tstamp;
+    uint8_t *p;
+
+    p = pktmbuf_mtod_offset(m, char *, sizeof(pktmbuf_t));
```
maryamtahhan, line 178: On the rx buffer side I've had to compensate for __get_mbuf_rx_aligned() subtracting the buf_headroom to the offset for the buffer.

If I do pktmbuf_mtod(m, char *); and hexdump the packet data it's all 0s... I'm wondering if you have encountered something similar with cnet?

```
 xskdev_buf_set_data(xi, *bufs, offset - xi->buf_mgmt.buf_headroom);
```
https://github.com/CloudNativeDataPlane/cndp/blob/main/lib/core/xskdev/xskdev.c#L307

KeithWiles: This does not appear to be right. The offset is the location of the packet data and subtracting buf_headroom seems wrong. I do not see this problem in CNET as nothing would work.

maryamtahhan: this is the only thing that prints out actual packet data at that offset. Vs all zeros from the 0 offset...

> I do not see this problem in CNET as nothing would work.

That's what I thought... so, I'm not sure if there's a setup issue with txgen that I missed... or if something has changed under the hood of xdpdev in the newer AF_XDP APIs... OR it's just something I'm missing.

The idea for the time stamping is pretty straight forward. On the TX side insert the timestamp at the start of the data part of the packet. On RX side, read the timestamp and calculate the latency.

if I just try to setup p as:
```
p = pktmbuf_mtod(m, char *);
```
and then hexdump the contents at that location - I see 0s for sizeof(pktmuf_t)...

If I offset as shown I'm reading the timestamp from the right place... so something is a miss...

KeithWiles: The latency timestamp needs to be placed at data_offset + sizeof(L2 header) + sizeof(L3 header) + sizeof(L4 header). As the data_offset should point to start of the packet unless it has been updated. Adding sizeof(pktmbuf_t) is only moving the pointer by 64 bytes.

maryamtahhan: > The latency timestamp needs to be placed at data_offset + sizeof(L2 header) + sizeof(L3 header) + sizeof(L4 header). As the data_offset should point to start of the packet unless it has been updated. Adding sizeof(pktmbuf_t) is only moving the pointer by 64 bytes.

That is where I'm putting it (and reading it from). At the `payload` part of the packet.

If I just look at the incoming buffer and print out the contents from data_offset there's a bunch of zeros and not the L2 header as I was expecting.

KeithWiles: One of the patches we added was to move the umem.frame_headroom by sizeof(pktmbuf_t) to account for VPP needing to add its sizeof(vlib_buffer_t) to skip overwriting the pktmbuf_t header. This change was in the VPP example plugin code.

Now the umem.frame_headroom is defaulted to sizeof(pktmbuf_t) as the offset. When the frame_headroom was zero all RX data offsets were placed at 256 bytes from the beginning of the UMEM buffer. Now the packet data on RX will start at 256 + sizeof(pktmbuf_t). Need to verify this is the case as I do not remember exactly.

For txgen it should be not updating the data_offset from the xskdev driver. Can you tell the exact offset from the beginning of the UMEM buffer the packet data is being placed?

KeithWiles: Did you have any luck on finding the problem here?

maryamtahhan: Will try to add more details in the morning. didn't get a chance to debug more today

KeithWiles: I grab your branch and put a hexdump in the txgen_tstamp_pointer_rx() routine and I am seeing L2 headers. I am not sending true latency packet, but regular UDP packets. I will look the transmit side real quick.

KeithWiles: I think line 178 in txgen.c inside the txgen_tstamp_pointer_rx()

p = pktmbuf_mtod_offset(m, char *, sizeof(pktmbuf_t));
should be
p = pktmbuf_mtod(m, char *);

A the offset zero is where the L2 header starts.

usrtools/txgen/app/cli-functions.c
```diff
@@ -276,14 +347,21 @@ pcap_cmd(int argc, char **argv)
...
+                                     {-1, NULL}};
 
 static const char *start_help[] = {
     "",
     "start <portlist>                   - Start transmitting packets",
     "stop <portlist>                    - Stop transmitting packets",
     "stp                                - Stop all lports from transmitting",
     "str                                - Start all lports transmitting",
+    "start <portlist> latsampler        - Start latency sampler, make sure to set sampling "
+    "parameters before starting",
```
KeithWiles, line 363: Does this line need to adjusted to be under the Start word?

usrtools/txgen/app/cli-functions.c
```diff
@@ -276,14 +347,21 @@ pcap_cmd(int argc, char **argv)
     return 0;
 }
 
-static struct cli_map start_map[] = {{10, "start %P"}, {20, "stop %P"}, {-1, NULL}};
+static struct cli_map start_map[] = {{10, "start %P"},
+                                     {20, "stop %P"},
+                                     {50, "start %P latsampler"},
```
KeithWiles, line 352: Could add %|latsampler|lat|latency if you like. Just makes it simpler to remember

usrtools/txgen/app/cli-functions.c
```diff
@@ -504,6 +588,40 @@ misc_cmd(int argc, char **argv)
     return 0;
 }
 
+static struct cli_map page_map[] = {{10, "page %d"},
+                                    {11, "page "
+                                         "%|main|pcap|latency"},
```
KeithWiles, line 593: This line should be on line 592. Could use 'lat' as a shorthand name.

usrtools/txgen/app/cmds.c
```diff
@@ -977,3 +983,168 @@ enable_capture(port_info_t *info, uint32_t state)
...
+ *
+ * RETURNS: N/A
+ *
+ * SEE ALSO:
+ */
+
+void
+txgen_start_latency_sampler(port_info_t *info)
+{
+
```
KeithWiles, line 1023: blank line needs to be removed.

usrtools/txgen/app/cmds.c
```diff
@@ -977,3 +983,168 @@ enable_capture(port_info_t *info, uint32_t state)
...
+{
+    if (state == ENABLE_STATE)
+        txgen_start_latency_sampler(info);
+    else if (state == DISABLE_STATE)
+        txgen_stop_latency_sampler(info);
+}
+
+/**
+ *
+ * txgen_start_latency_sampler - Starts latency sampler.
```
KeithWiles, line 1010: The comments for this API and start_stop_latency_sampler are same, should they be?

usrtools/txgen/app/cmds.c
```diff
@@ -977,3 +983,168 @@ enable_capture(port_info_t *info, uint32_t state)
...
+        return;
+    }
+
+    info->latsamp_stats.pkt_counter = 0;
+    info->latsamp_stats.next        = 0;
+    info->latsamp_stats.idx         = 0;
+    info->latsamp_stats.num_samples = info->latsamp_num_samples;
+
+    if (info->pkt.pktSize < (CNE_ETHER_MIN_LEN - ETHER_CRC_LEN) + sizeof(tstamp_t))
+        info->pkt.pktSize += sizeof(tstamp_t);
```
KeithWiles, line 1040: info->pkt.pktsize could be much smaller than (CNE_ETHER_MIN_LEN -ETHER_CRC_LEN) + sizeof(tstamp_t), if true we only add sizeof(tstamp_t), may be need to make sure is set to exactly the above length.

usrtools/txgen/app/cmds.c
```diff
@@ -977,3 +983,168 @@ enable_capture(port_info_t *info, uint32_t state)
...
+    }
+
+    /* Reset stats data */
+    info->latsamp_stats.pkt_counter = 0;
+    info->latsamp_stats.next        = 0;
+    info->latsamp_stats.idx         = 0;
+    info->latsamp_stats.num_samples = 0;
+
+    if (info->pkt.pktSize >= (CNE_ETHER_MIN_LEN - ETHER_CRC_LEN) + sizeof(tstamp_t))
+        info->pkt.pktSize -= sizeof(tstamp_t);
```
KeithWiles, line 1090: Same here as above?

usrtools/txgen/app/cmds.c
```diff
@@ -977,3 +983,168 @@ enable_capture(port_info_t *info, uint32_t state)
...
+
+    info->pkt.ipProto = IPPROTO_UDP;
```
KeithWiles, line 1092: When stopping latency why do we set it back to IPPROTO_UDP, is it something else? If we are forcing UDP and TCP is not allowed it seems strange to me. I did not write the latency code at least I do not remember, but I should have flagged this in the original code. :-(

usrtools/txgen/app/cmds.c
```diff
@@ -977,3 +983,168 @@ enable_capture(port_info_t *info, uint32_t state)
...
+        return;
+
+    if ((str[0] >= '0') && (str[0] <= '9')) {
+        page = atoi(str);
+        if (page > pktdev_port_count())
+            return;
+    }
+
+    /* Switch to the correct page */
+    if (_cp("latency") || _cp("lat")) {
```
KeithWiles, line 1123: ah, so we do need to add lat as a shorthand for page command.

usrtools/txgen/app/txgen.c
```diff
@@ -503,20 +644,24 @@ txgen_main_receive(port_info_t *info __cne_unused, pktmbuf_t *pkts_burst[], uint
     /*
      * Read packet from RX queues and free the mbufs
      */
-    nb_rx = pktdev_rx_burst(lport->lpid, pkts_burst, nb_pkts);
-    if ((nb_rx == 0) || (nb_rx == PKTDEV_ADMIN_STATE_DOWN))
+    nb_rx = pktdev_rx_burst(lport->lpid, (pktmbuf_t **)pkts_burst, nb_pkts);
+    if (nb_rx == PKTDEV_ADMIN_STATE_DOWN) {
```
KeithWiles, line 648: You can remove the single line if statement {}

KeithWiles (changes requested):
Only a few minor nits. Thanks

usrtools/txgen/app/txgen.c
maryamtahhan: > cne_hexdump(NULL, pktmbuf_mtod(pkt, char *), 16);

Hey Keith
I did a quick test this morning, setup a UDP packet and sent it through. I added some instrumentation to _process_tstamp() below:

```
static __inline__ void
_process_tstamp(port_info_t *info, pktmbuf_t *pkt)
{
    uint64_t lat, jitter;
    tstamp_t *tstamp;

	cne_printf("pktmbuf_mtod(pkt, char *)\n");
	cne_hexdump(NULL, "pktmbuf_mtod(pkt, char *)", pktmbuf_mtod(pkt, char *), 16);
	cne_printf("pktmbuf_mtod_offset(pkt, char *, sizeof(pktmbuf_t)\n");
	cne_hexdump(NULL, "pktmbuf_mtod_offset(pkt, char *, sizeof(pktmbuf_t)", pktmbuf_mtod_offset(pkt, char *, sizeof(pktmbuf_t)), 16);

    tstamp = txgen_tstamp_pointer_rx(info, pkt);
    if (tstamp->magic == TSTAMP_MAGIC) {
        lat                    = (cne_rdtsc_precise() - tstamp->timestamp);
        latsamp_stats_t *stats = &info->latsamp_stats;
        uint64_t now           = cne_rdtsc_precise();

        info->avg_latency += lat;
        if (lat > info->prev_latency)
            jitter = lat - info->prev_latency;
        else
            jitter = info->prev_latency - lat;
        if (lat > info->max_latency)
            info->max_latency = lat;
        if (jitter > info->jitter_threshold_clks)
            info->jitter_count++;
        info->prev_latency = lat;

        stats->pkt_counter++;
        if (stats->next == 0 || now >= stats->next) {
            if (stats->idx < stats->num_samples) {
                stats->data[stats->idx] =
                    (lat * Billion) / cne_get_timer_hz(); /* Do we want to keep it as cycles? */
                stats->idx++;
            }

            /* Calculate next sampling point TODO: Use poisson */
            if (info->latsamp_type == LATSAMPLER_POISSON) {
                double next_possion_time_ns = next_poisson_time(info->latsamp_rate);
                stats->next =
                    now + next_possion_time_ns * (double)cne_get_timer_hz();        // Time based
            } else {        // LATSAMPLER_SIMPLE or LATSAMPLER_UNSPEC
                stats->next = now + cne_get_timer_hz() / info->latsamp_rate;        // Time based
            }
        }
    } else {
        info->magic_errors++;
    }

    info->latency_nb_pkts++;
}

```

I only see the L2 headers when I'm using pktmbuf_mtod_offset(pkt, char *, sizeof(pktmbuf_t)

```
TXGen:/> pktmbuf_mtod(pkt, char *)
 at [0x7f9caec000c0], len=16
00000000: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 | ................
pktmbuf_mtod_offset(pkt, char *, sizeof(pktmbuf_t)
 at [0x7f9caec00100], len=16
00000000: EC F4 BB C0 B6 2A EC F4 BB C0 B6 28 08 00 45 00 | .....*.....(..E.
```

What version of libbpf are you using?

I'm using libxdp and I'm wondering if something changed under the hood. I can try to use the same version of libbpf as you to see if I can duplicate the behaviour you are seeing (L2 headers at offset 0)?

usrtools/txgen/app/cli-functions.c
maryamtahhan: will do

usrtools/txgen/app/cli-functions.c
maryamtahhan: I will fix it

usrtools/txgen/app/txgen.c
maryamtahhan: these are my bpf and xdp lib versions:

```
Run-time dependency libxdp found: YES 1.2.5
Run-time dependency libbpf found: YES 0.8.0
```

KeithWiles: I am using Run-time dependency libbpf found: YES 0.6.0, but I need to move to libxdp. Any libbpf above 6.1 will cause a set of warning as the deprecated the xsk APIs.

maryamtahhan: let me bounce back to 0.6.0 and see what the behaviour is there with the offset weirdness...

maryamtahhan: so I tried it with 0.6.0 and I see the L2 header with  pktmbuf_mtod(pkt, char *)... so it looks like something did change under the hood. I don't know if this is a bug we want to flag to Magnus, but for now I will submit a patch that will fix the issue for CNDP applications using libxdp 

```
TXGen:/> pktmbuf_mtod(pkt, char *)
 at [0x7f833ec01900], len=16
00000000: 02 A7 A2 BC 51 30 EC F4 BB C0 B6 2A 08 00 45 00 | ....Q0.....*..E.
pktmbuf_mtod_offset(pkt, char *, sizeof(pktmbuf_t)
 at [0x7f833ec01940], len=16
00000000: 61 62 63 64 65 66 67 68 69 6A 6B 6C 6D 6E 6F 70 | abcdefghijklmnop

```

KeithWiles: Because the APIs in libbpf 0.8.0 are deprecated it maybe they did not go back and fix something they added in libxdp.

usrtools/txgen/app/cmds.c
maryamtahhan: I see: 
```
start_stop_latency_sampler - Starts or stops latency sampler.
```
while this is

```
txgen_start_latency_sampler - Starts latency sampler.
```

usrtools/txgen/app/cmds.c
maryamtahhan: TBH I think I can keep track of the original packet size and reset it to this rather than trying to reset it in this way...

usrtools/txgen/app/cmds.c
maryamtahhan: yeah will fix

usrtools/txgen/app/cmds.c
maryamtahhan: No worries - makes sense. I will clean it up

usrtools/txgen/app/txgen.c
KeithWiles: We even have this noted in the INSTALL.md file and it does fix the build issue. :-(

usrtools/txgen/app/cli-functions.c
maryamtahhan: it looks like clang format doesn't like this:

```
--- ./usrtools/txgen/app/cli-functions.c	(original)
+++ ./usrtools/txgen/app/cli-functions.c	(reformatted)
@@ -360,7 +360,7 @@
     "stp                                      - Stop all lports from transmitting",
     "str                                      - Start all lports transmitting",
     "start <portlist> latsampler|lat|latency  - Start latency sampler, make sure to set sampling "
-                                                "parameters before starting",
+    "parameters before starting",
     "stop <portlist> latsampler|lat|latency   - Stop latency sampler, dumps to file if specified",
     CLI_HELP_PAUSE,
     NULL};
```

KeithWiles (approved):
Thanks Maryam, look good!

KeithWiles:
@maryamtahhan this one has a conflict in usertools/txgen/app/meson.build the deps line has the conflict. I looked at the conflict and it appears this is the correct line.

deps += [cli, kvargs, pktdev, pktmbuf, mempool, mmap, ring, jcfg, thread, uds, metrics, timer, pmd_af_xdp, pmd_tap, utils]

The order is different and timer/utils are different from the af_packet I believe.

maryamtahhan:
just rebased manually there - hopefully that sorts out the merging blockage

KeithWiles (approved):

KeithWiles:
Thanks that fixed the problem.

PR #189, opened 2022-09-07
xskdev: revert b98a35d.

maryamtahhan:
This reverts xskdev changes in commit b98a35d. Tested with libxdp:
1.2.4, 1.2.5, 1.2.6. This is not an issue for cnet or txgen.

I've done quite a bit of testing and I can't find the combination of libraries that was causing the issue.


KeithWiles (approved):
Thank you

PR #192, opened 2022-09-10
fix up performance drop for FQ

KeithWiles:
The FQ routine fq_add() was allocating mbufs and freeing them when the FQ was full, this caused a performance drop. The code will now determine if the FQ needs mbufs and how many before allocating the mbufs from the pktmbuf pool.

During testing of tx only cndpfwd and the idlemgr was enabled causing the thread to sleep due to no RX traffic. Modified the txonly routine to return 1 to make it look like the port was active to idlemgr.

### DIFF
--- a/examples/cndpfwd/main.c
+++ b/examples/cndpfwd/main.c
@@ -225,7 +225,7 @@ _txonly_test(jcfg_lport_t *lport, struct fwd_info *fwd)
         pd->tx_overrun += n;
     }
 
-    return 0;
+    return n_pkts;
 }
 
 static int
@@ -288,7 +288,7 @@ _txonly_rx_test(jcfg_lport_t *lport, struct fwd_info *fwd)
         pd->tx_overrun += n;
     }
 
-    return 0;
+    return n_pkts;
 }
 
 static void
--- a/lib/core/xskdev/xskdev.c
+++ b/lib/core/xskdev/xskdev.c
@@ -191,82 +191,50 @@ err_prefer:
     return 0;
 }
 
-static int
-fq_reserved(xskdev_info_t *xi, uint16_t size)
+static inline void
+xsk_ring_prod__cancel(struct xsk_ring_prod *prod, __u32 nb)
 {
-    struct xskdev_umem *ux   = xi->rxq.ux;
-    struct xsk_ring_prod *fq = &ux->fq;
-    void *bufs[size];
-    uint16_t nb_bufs = size;
-    uint32_t pos     = 0;
-    int ret          = 0;
-
-    ret = xsk_ring_prod__reserve(fq, nb_bufs, &pos);
-    if (ret != nb_bufs)
-        return -1;
-
-    if (xskdev_buf_alloc(xi, bufs, nb_bufs) <= 0) {
-        xi->stats.fq_alloc_failed++;
-        return -1;
-    }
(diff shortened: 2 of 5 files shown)


KeithWiles:
@manojgop can you have a look at this patch and see if it works for you? For some reason I can not add you as a reviewer.

manojgop:
> @manojgop can you have a look at this patch and see if it works for you? For some reason I can not add you as a reviewer.

I checked this PR with cndpfwd application in my setup using DPDK pktgen as packet generator. loopback performance is same before/after applying this patch

examples/cndpfwd/main.c
```diff
@@ -225,7 +225,7 @@ _txonly_test(jcfg_lport_t *lport, struct fwd_info *fwd)
         pd->tx_overrun += n;
     }
 
-    return 0;
+    return 1;
```
manojgop, line 228: Can we return the number of packets send by `_txonly_test()` instead of 1 ?

lib/core/xskdev/xskdev.c
```diff
@@ -191,82 +191,50 @@ configure_busy_poll(xskdev_info_t *xi)
...
+            xi->stats.fq_reserve_failed++;
+            break;
+        }
 
-    for (int i = 0; i < nb; i++) {
-        void *buf       = bufs[i];
-        uint64_t offset = (uint64_t)xskdev_buf_get_addr(xi, buf) - (uint64_t)ux->umem_addr -
-                          (uint64_t)xi->buf_mgmt.pool_header_sz;
+        nb_bufs = xskdev_buf_alloc(xi, (void **)bufs, FQ_ADD_BURST_COUNT);
+        if (nb_bufs != FQ_ADD_BURST_COUNT) {
```
manojgop, line 218: So does this mean, if we try to allocate FQ_ADD_BURST_COUNT and we can't allocate all buffers  , cancel the buffers which got allocated by `xskdev_buf_alloc()`?  Is there a way to check the size of FQ to see if it can accommodate at least FQ_ADD_BURST_COUNT buffer before calling `xskdev_buf_alloc()` ?

KeithWiles:
@manojgop Can you approve the PR?

manojgop (approved):

lib/core/xskdev/xskdev.c
KeithWiles: When we do the xsk_ring_prod__reserver() call we check to make sure we can add the buffers to the ring. Is this what you meant?

manojgop: ok. Does xsk_ring_prod__reserver() ensures we can allocate at least FQ_ADD_BURST_COUNT buffers ?

manojgop (approved):

manojgop:
> @manojgop Can you approve the PR?

approved

lib/core/xskdev/xskdev.c
KeithWiles: The` xsk_ring_prod__reserve()` call above the allocation will insure we can put `FQ_ADD_BURST_COUNT` on the FQ ring. The `xskdev_buf_alloc()` call is checked to make sure we get `FQ_ADD_BURST_COUNT` of mbufs.

KeithWiles:
Thanks

PR #193, opened 2022-09-12
Use CC build crate for building Rust bindings.

manojgop:
CC build crate allows to pass custom cndp install path when building Rust bindings. Meson doesn't allow to read CNDP_INSTALL_PATH env variable.


eli-schwartz:
Does it really need to be an environment variable?

If you have existing code handling that environment variable and calling Meson, can't you just set that up as $PKG_CONFIG_PATH?

I guess it's also possible to just add a meson_options.txt option such that `"-Dcndp_install_path={}", cndp_install_path` can be passed to Meson.

manojgop:
> Does it really need to be an environment variable?
> 
> If you have existing code handling that environment variable and calling Meson, can't you just set that up as $PKG_CONFIG_PATH?
> 
> I guess it's also possible to just add a meson_options.txt option such that `"-Dcndp_install_path={}", cndp_install_path` can be passed to Meson.

Existing [build.rs](https://github.com/CloudNativeDataPlane/cndp/blob/main/lang/rs/bindings/cne-sys/build.rs#L56) was using [Rust Meson crate](https://docs.rs/meson/1.0.0/meson/fn.build.html). But Meson crate doesn't allow to pass any user defined arguments or build context (other than project,  build output directory). [Rust CC crate](https://docs.rs/cc/latest/cc/struct.Build.html) allows to pass the include directory as an argument to build context. So it was easier to build c source file in CNDP Rust bindings code by passing custom include path, using CC crate than Meson crate.

eli-schwartz:
I would say that the crate itself is flawed and should be fixed, but apparently the author of that crate has only ever interacted with github once, to create a single repository with one commit and release that as meson-1.0.0 on crates.io

It seems likely you can do a lot better.

Maybe the Meson crate should be forked and improved, and released as a new, better project?

manojgop:
> I would say that the crate itself is flawed and should be fixed, but apparently the author of that crate has only ever interacted with github once, to create a single repository with one commit and release that as meson-1.0.0 on crates.io
> 
> It seems likely you can do a lot better.
> 
> Maybe the Meson crate should be forked and improved, and released as a new, better project?

Yes. This Rust Meson crate is probably not maintained anymore. I couldn't find the corresponding source code repository for this. The existing [source](https://docs.rs/crate/meson/1.0.0/source/src/lib.rs) looks very simple which invokes meson/ninja command. Do you know if there are any alternate Rust crates for Meson or anything under works ? Modifying existing Rust meson crate is not in scope of this PR. Do you have any concerns in using Rust CC crate for CNDP Rust bindings ? If not, could we merge this PR and revisit using meson later ?

KeithWiles (approved):

PR #194, opened 2022-09-14
Fixes for Fedora 36 build

donaldh:
These commits fix issues I encountered when building on Fedora 36, with gcc 12.1.1.


KeithWiles (approved):
Thank you for the patches.

We need signed-off-by lines in the commits to be able to merge these changes.

PR #196, opened 2022-09-15
Enable lto for Rust build which improves performance.

manojgop:

### DIFF
--- a/lang/rs/Cargo.toml
+++ b/lang/rs/Cargo.toml
@@ -11,3 +11,6 @@ members = [
     "examples/fwd"
 ]
 
+[profile.release]
+lto = true
+


lang/rs/Cargo.toml
```diff
@@ -11,3 +11,6 @@ members = [
     "examples/fwd"
 ]
 
+[profile.release]
+lto = true
```
KeithWiles, line 15: What is the performance gain when using this feature?

manojgop: I've noticed 10-15% drop in performance if lto is not enabled with Rust builds.

KeithWiles (approved):

PR #197, opened 2022-09-15
xskdev:force zero copy mode

KeithWiles:
When the driver did not support zero-copy mode it would default to copy mode, but we did not force zero-copy mode. The performance would drop and we did not report an error message to alert the operator.

Force zero copy mode unless asked to do copy mode i.e., skb_mode.


jeffreybshaw:
We want the kernel to pick the best mode, right? Can you test with `veth` driver and make sure you don't need to specify skb_mode, and the socket can still be created?

PR #199, opened 2022-09-16
Remove LD_PRELOAD of pmd library for Rust applications

manojgop:
- Use "whole-archive" to link pmd_af_xdp library and avoid LD_PRELOAD.
- Ignore cargo clippy lint for generated bindings file.
- Fix lint in Rust CNE.

### DIFF
--- a/lang/rs/bindings/cne-sys/build.rs
+++ b/lang/rs/bindings/cne-sys/build.rs
@@ -53,6 +53,11 @@ fn main() {
     println!("cargo:rustc-link-search=native={}", libcndp_so_path);
     println!("cargo:rustc-link-lib=cndp");
 
+    // CNDP pmd_af_xdp static library should be linked as whole archive.
+    // This will avoid preload PMD (eg: LD_PRELOAD=libpmd_af_xdp.so) while running
+    // CNDP CNE Rust binaries.
+    println!("cargo:rustc-link-lib=static:-bundle,+whole-archive=pmd_af_xdp");
+
     // Tell cargo to invalidate the built crate whenever the .h or .c or meson.build files in bindings directory changes.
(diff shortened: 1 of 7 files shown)


lang/rs/bindings/cne-sys/build.rs
```diff
@@ -53,6 +53,11 @@ fn main() {
     println!("cargo:rustc-link-search=native={}", libcndp_so_path);
     println!("cargo:rustc-link-lib=cndp");
 
+    // CNDP pmd_af_xdp static library should be linked as whole archive.
+    // This will avoid preload PMD (eg: LD_PRELOAD=libpmd_af_xdp.so) while running
+    // CNDP CNE Rust binaries.
+    println!("cargo:rustc-link-lib=static:-bundle,+whole-archive=pmd_af_xdp");
```
KeithWiles, line 59: When using --whole-archive for pmd_af_xdp and you stated this is a static library, why do we need to build shared libraries or was the static text a typo?

manojgop: whole-archive is linked as static only for pmd_af_xdp. Rest of the CNDP CNE libraries are linked as shared library. So only pmd_af_xdp will be part of Rust binary. That way the binary size can be kept small.

KeithWiles (approved):
As a longer term we need to figure out a cleaner why to include PMDs. Rust example only needs the af_xdp pmd, but we need to solve the problem when Rust, Go or C code needs other PMDs.

manojgop:
> As a longer term we need to figure out a cleaner why to include PMDs. Rust example only needs the af_xdp pmd, but we need to solve the problem when Rust, Go or C code needs other PMDs.

One option is we can combine all pmds into one static library or else need to link each pmd needed as whole-archive

PR #201, opened 2022-09-16
xskdev:FQ needs to be kept full

KeithWiles:
The FQ needs to be kept full, but the current method to keep it full created a performance drop in the testing trends. The previous fix attempted to keep the FQ full, but was allocating mbufs and freeing them when not able to push the mbufs top the FQ ring. This fix checks to determine if the FQ can be filled with FQ_ADD_MBUF_COUNT number of mbufs before allocating the mbufs first. The fq_add() has a new argument to attempt to add mbufs a number of times.


KeithWiles:
Applying to see if the performance changes tonight.

PR #208, opened 2022-09-22
Fix static build: Generate static library for pmds.

manojgop:
pmds uses constructor which runs before main function to register pmds. linker doesn't include any symbols from PMDs in the application binary if application doesn't explicitly use any symbols from PMD library. CNDP applications need not directly call any functions in pmd library and pmd library symbols is not included in application by linker. So pmds should be linked as static whole-archive. 

Remove pmds libraries from libcndp.so and libcndp.a.
Generate libcndp_pmds.a which groups all PMD libraries.

Build works in both cases: "make rebuild" and "make static_build=1 rebuild".

Note: shared libraries are still generated for other CNDP CNE libraries in default non static build. For PMDs, generate only static libraries.

### DIFF
--- a/lib/core/pmds/net/af_packet/meson.build
+++ b/lib/core/pmds/net/af_packet/meson.build
@@ -6,7 +6,8 @@ headers = files('pmd_af_packet.h')
 
 deps += [cne, mempool, mmap, pktdev, pktmbuf]
 
-libpmd_af_packet = both_libraries('pmd_af_packet', sources, install: true, dependencies: deps)
+libpmd_af_packet = static_library('pmd_af_packet', sources, install: true, dependencies: deps)
+
 pmd_af_packet = declare_dependency(link_with: libpmd_af_packet, include_directories: include_directories('.'))
 
 cndp_pmds += libpmd_af_packet
--- a/meson.build
+++ b/meson.build
@@ -31,6 +31,7 @@ cndp_libs = []
 extra_ldflags = []
 enabled_libs = []
 cndp_pmds = []
+enabled_pmd_libs = []
 
 cne_conf = configuration_data()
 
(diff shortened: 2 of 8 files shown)


lib/core/pmds/net/af_packet/meson.build
```diff
@@ -6,7 +6,12 @@ headers = files('pmd_af_packet.h')
 
 deps += [cne, mempool, mmap, pktdev, pktmbuf]
 
-libpmd_af_packet = both_libraries('pmd_af_packet', sources, install: true, dependencies: deps)
+if use_static_libs
+  libpmd_af_packet = static_library('pmd_af_packet', sources, install: true, dependencies: deps)
+else
+  libpmd_af_packet = both_libraries('pmd_af_packet', sources, install: true, dependencies: deps)
```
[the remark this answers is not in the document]
manojgop: Another option is we can use library() for pmds and we can remove libpmd_xxx from libcndp.so. And let applications link with libcndp and libpms separately. In fact, IMO that's better solution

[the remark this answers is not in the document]
KeithWiles: The meson.build should only build and create the libcndp.so when we are not building static libs. Need to update the meson.build to only create libcndp.so or libcndp.a

```c
cndp_a_name = 'libcndp.a'
cndp_so_name = 'libcndp.so'

mklib = find_program('tools/mklib.sh')
if use_static_libs
    run_command(mklib, cndp_a_name, libcndp_a, check: true)
    cndp_a = custom_target('libcndp_a_target',
        output: cndp_a_name,
        input: join_paths('/tmp', cndp_a_name),
        command:[find_program('cp'), '@INPUT@', cndp_a_name],
        install_dir: join_paths('lib', 'x86_64-linux-gnu'),
        install: true)
else
    run_command(mklib, cndp_so_name, libcndp_so, check: true)
    cndp_so = custom_target('libcndp_so_target',
        output: cndp_so_name,
        input: join_paths('/tmp', cndp_so_name),
        command:[find_program('cp'), '@INPUT@', cndp_so_name],
        install_dir: join_paths('lib', 'x86_64-linux-gnu'),
        install: true)
endif
```

I think this needs to be done, but I am not convinced the PR code is the best solution.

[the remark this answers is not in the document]
manojgop: IMO we need only static for pmds. We can remove libpmd_xx.so (shared pmd library) from libcndp.so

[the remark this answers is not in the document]
KeithWiles: Just saw your response and it sounds more reasonable. Please update the top level meson.build to only build libcndp.so or libcndp.a when needed and do your suggested solution.

meson.build
```diff
@@ -429,21 +429,24 @@ cndp_a_name = 'libcndp.a'
...
+        install: true)
+else
+    run_command(mklib, cndp_so_name, libcndp_so, check: true)
+    cndp_so = custom_target('libcndp_so_target',
+        output: cndp_so_name,
+        input: join_paths('/tmp', cndp_so_name),
+        command:[find_program('cp'), '@INPUT@', cndp_so_name],
+        install_dir: join_paths('lib', 'x86_64-linux-gnu'),
+        install: true)
+endif
```
KeithWiles, line 449: Did we need to create a libcndp_pmds.a file similar to libcndp.so/.a ?

manojgop: Good point. But I think mostly applications doesn't need to use all pmds. So linking all pmds as whole-archive (using libcndp_pmd.a) is unnecessary IMO. So would selecting liking of required PMDs as whole-archive based on application need seems better.

KeithWiles: We have so few PMDs it would only add about 64K to the image.

KeithWiles: yes, please create it.

manojgop: Is tun also a pmd that needs to be included in `libcndp_pmds.a` ?

KeithWiles: No Tun is how we allocate the interfaces, but it is called out for the PMD and it is not a PMD

manojgop: Updated PR

KeithWiles (approved):
Thanks

PR #211, opened 2022-09-23
fix go executables using static pmds

KeithWiles:
The go code examples and tests were failing because of the PMDs are now static libraries. The problem turns out that building cgo code with --whole-archive is not allowed unless you enable the CGO_LDFLAGS_ALLOW environment variable before building.

Also a number of the binding go files included CGO options multiple times, which is not required. We only need to have the CGO options once per package. Moved the CGO directives into a single file cgo_config.go then removed these lines from the rest of the files.

Make sure the TUN library is included in the libcndp.so/.a files and remove it from the libcndp_pmds.a file. The lib/core/pmds/tun library is not a PMD.

Tested the test routines and the examples/fwd code as shared objects. Turns out having static go binaries is complex with CGO and the go test command will not accept static linking.

### DIFF
--- a/lib/core/pmds/tun/meson.build
+++ b/lib/core/pmds/tun/meson.build
@@ -9,4 +9,5 @@ deps += [include]
 libtun = library('tun', sources, install: true, dependencies: deps)
 tun = declare_dependency(link_with: libtun, include_directories: include_directories('.'))
 
+enabled_libs += 'tun'
 cndp_libs += libtun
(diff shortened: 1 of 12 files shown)


jeffreybshaw (approved):

lib/core/pmds/tun/meson.build
```diff
@@ -9,4 +9,5 @@ deps += [include]
 libtun = library('tun', sources, install: true, dependencies: deps)
 tun = declare_dependency(link_with: libtun, include_directories: include_directories('.'))
 
+enabled_libs += 'tun'
```
manojgop, line 12: Can we keep tun directory outside pmds directory since tun is not a pmd ?

PR #213, opened 2022-09-30
txgen:fix static PMD builds

KeithWiles:
Make sure txgen and test-cne are built with the new static PMDs.


KeithWiles:
When I run the verify function test it fails, but If I run it locally then it works. Not sure what is happening now. I believe I have fixed up the library issues caused by converting the PMD libs into static only libraries. :-(

Issue #234, opened 2022-10-20
Rework the test-cne testing infrastructure


KeithWiles:
**Difficulty**: Easy

Enhance the test-cne infrastructure to be more consistent or easier to use. Need all of the tests to look similar and produce similar output.

For [example](https://github.com/CloudNativeDataPlane/cndp/blob/main/test/testcne/graph_test.c) to show this test is different then the others. This test uses printf and not the tst_info() or tst_error() type APIs, these need to be cleaned up to use the correct APIs from the tst_ XXX library.

byron-marohn:
Oof, didn't notice that pushing commits to a fork that referenced this number would leave comments here. I think we're done generating noise here for now. 

We've addressed several of the improvements suggested here for test files, but there are still some that need doing which someone else is welcome to work on.

KeithWiles:
I think the major changes have been done and we can open a new issue later to address any other issues.

PR #244, opened 2022-12-15
Update vector test to a consistent standard

byron-marohn:
Improves #234

- Use tst_info() instead of cne_printf()
- Print more clear test results
- Fix test function return value

### DIFF
--- a/test/testcne/vec_test.c
+++ b/test/testcne/vec_test.c
@@ -34,7 +34,7 @@ static struct test_data {
 };
 // clang-format on
 
-static bool
+static int
 test1(struct test_data *t)
 {
     int d, *u = NULL, ret = TST_PASSED;
@@ -43,7 +43,7 @@ test1(struct test_data *t)
     if (!u)
         return -1;
 
-    cne_printf("%s:\n", t->name);
+    tst_info("%s:\n", t->name);
 
     for (int j = 0; j < t->cnt; j++)
         vec_add(u, j + t->cnt);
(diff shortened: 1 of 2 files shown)


maryamtahhan (approved):

test/testcne/vec_test.c
```diff
@@ -135,6 +135,22 @@ test3(struct test_data *t)
     return ret;
 }
 
+static void
+result_print(tst_info_t *tst, int ret)
+{
+    if (tst == NULL) {
```
KeithWiles, line 141: We normally do not want {} around single line statements for  if, while, for, ... 

When you get a chance it would be great to fix these.

test/testcne/vec_test.c
```diff
@@ -135,6 +135,22 @@ test3(struct test_data *t)
...
+        tst_error("test case struct was empty");
```
KeithWiles, line 142: We should add a return statement here as tst_error() does not exit or return out of this function or restructure this into some type of if else like statement.

The tst->name later will cause an exception.

KeithWiles (commented):
The changes look good, but we need to fix some issues.

Thanks for the patch.

test/testcne/vec_test.c
maryamtahhan: Sorry - this is my fault - I forgot clang-format doesn't automatically catch these and I missed this :(

byron-marohn:
Will fix comments in a follow-on PR today, thanks for the review.

byron-marohn:
Improvements are in PR #248

PR #250, opened 2022-12-17
graph_test and graph_perf_test improvements

byron-marohn:
- Improves #234
- Replaced cne_printf() with tst_*()
- Improved output readability somewhat

### DIFF
--- a/test/testcne/graph_test.c
+++ b/test/testcne/graph_test.c
@@ -208,15 +208,13 @@ test_node0_worker(struct cne_graph *graph, struct cne_node *node, void **objs, u
         for (i = 0; i < nb_objs; i++) {
             data = (pktmbuf_t *)objs[i];
             if ((pktmbuf_udata64(data) >> 32) != tm->test_node[0].idx) {
-                cne_printf("Data idx miss match at node 0, expected"
-                           " = %u got = %u\n",
-                           tm->test_node[0].idx, (uint32_t)(pktmbuf_udata64(data) >> 32));
+                tst_error("Data idx miss match at node 0, expected = %u got = %u",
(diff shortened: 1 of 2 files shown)


test/testcne/graph_test.c
```diff
@@ -552,23 +545,23 @@ test_update_edges(void)
     count   = cne_node_edge_update(node_id, 0, tm->test_node[0].node.next_nodes,
                                    tm->test_node[0].node.nb_edges);
```
byron-marohn, line 546: Unfortunately this line throws an error in clang-format that we weren't able to fix. Attempting to follow its advice just results in a new error to change it back to this format. 

Since this line was not actually part of the changes in this pull request we wanted to move ahead anyway after spending entirely too much time trying to figure out how to fix it.

byron-marohn: We separated out this commit from the other test improvements so that it was more obvious that the clang format error is just due to this commit.

KeithWiles:
Which version of clang-format are you using. clang-format --version

I am using 15.0.2-1

byron-marohn:
This particular problem manifests not client side but on whatever server-side clang-format that runs within GitHub Actions. We edited the code manually in both directions. I'm not sure which version is run in GitHub actions but I can dig into that if you'd like me to.

KeithWiles:
Best I can tell it is clang-format version 13, at least the github action reports version 13

KeithWiles:
I created a new pull-request updating clang-format to version 15.0.2. If you could have a look and see if that fixes the problem plus approve the PR it would be great. #255

KeithWiles:
Looks like clang-format passes now.

KeithWiles (approved):
LGTM thank you

PR #252, opened 2022-12-17
Fix build /tmp conflict on multi-user system

byron-marohn:
This fixes a problem where two files are placed in /tmp that are owned by the user. The next user that tries to build on the system fails, because those files aren't writable to them.

This patch puts the /tmp files in a separate randomly generated temp directory in /tmp to avoid this conflict.


byron-marohn:
We're definitely expecting comments/feedback on this approach. This was an issue we hit during testing - only one of us was able to build the project at a time on a shared box we set up to make development easier. 

One thing to note - the temp directory isn't cleaned up later. It only contains about 1k worth of data, but with hundreds of builds eventually there will be hundreds of `/tmp/tmp.SZ1SdLEcT8/` directories. Doesn't seem to be a straightforward way to delete these, open to ideas.

meson.build
```diff
@@ -456,10 +457,10 @@ foreach lib:enabled_pmd_libs
...
 cndp_pmds_a_name = 'libcndp_pmds.a'
-run_command(mklib, cndp_pmds_a_name, libcndp_pmds_a, check: true)
+run_command(mklib, tempdir, cndp_pmds_a_name, libcndp_pmds_a, check: true)
 cndp_pmds_a = custom_target('libcndp_pmds_a_target',
     output: cndp_pmds_a_name,
-    input: join_paths('/tmp', cndp_pmds_a_name),
+    input: join_paths(tempdir, cndp_pmds_a_name),
     command:[find_program('cp'), '@INPUT@', cndp_pmds_a_name],
     install_dir: join_paths('lib', 'x86_64-linux-gnu'),
     install: true)
```
KeithWiles, line 466: If we add a run_command() after this custom_target() to remove the temp directory will that work?

KeithWiles, line 466: We should be able to fix this problem with the following:

```
mklib = find_program('tools/mklib.sh')
build_dir = meson.current_build_dir()

if use_static_libs
    cndp_a = custom_target('libcndp_a_target',
        output: 'libcndp.a',
        command:[mklib, build_dir, '@OUTPUT@', libcndp_a],
        install_dir: join_paths('lib', 'x86_64-linux-gnu'),
        install: true)
else
    cndp_so = custom_target('libcndp_so_target',
        output: 'libcndp.so',
        command:[mklib, build_dir, '@OUTPUT@', libcndp_so],
        install_dir: join_paths('lib', 'x86_64-linux-gnu'),
        install: true)
endif

# Generate libcndp_pmds.a
libcndp_pmds_a = []
foreach lib:enabled_pmd_libs
    libcndp_pmds_a += 'lib' + lib + '.a '
endforeach

cndp_pmds_a = custom_target('libcndp_pmds_a_target',
    output: 'libcndp_pmds.a',
    command:[mklib, build_dir, '@OUTPUT@', libcndp_pmds_a],
    install_dir: join_paths('lib', 'x86_64-linux-gnu'),
    install: true)
```

Building the linker files directly into the build directory to avoid the /tmp conflict.

byron-marohn:
Thanks for the suggestion. Will try that out. Moving slowly here, might take a little more time to get back to this.

KeithWiles (approved):
LGTM, Thank you.

Issue #231, opened 2022-10-20
Add L3 Fowarding to cndpfwd application


KeithWiles:
**Difficulty**: Medium

Add L3 forwarding to [cndpfwd](https://github.com/CloudNativeDataPlane/cndp/tree/main/examples/cndpfwd) using [RIB](https://github.com/CloudNativeDataPlane/cndp/tree/main/lib/usr/clib/rib)/[FIB](https://github.com/CloudNativeDataPlane/cndp/tree/main/lib/usr/clib/fib) APIs.

KeithWiles:
Thanks, we can close this issue now that the PR has been merged.

PR #254, opened 2022-12-17
Add L3 forwarding to cndpfwd example

stvml:
This commit enhances the cndpfwd example application by adding a new `l3-fwd` mode. In this mode, incoming packets are forwarded based on a FIB defined in `fwd.jsonc`, in the `options:l3fwd-fib-rules` section. Each FIB entry consists of an IP/prefix to match, a destination MAC for the next hop, and a TX port to use.

Targets #231.

### DIFF
--- a/examples/cndpfwd/l3-fwd.c
+++ b/examples/cndpfwd/l3-fwd.c
@@ -1,0 +1,112 @@
+/* SPDX-License-Identifier: BSD-3-Clause
+ * Copyright (c) 2022 Intel Corporation.
+ */
+
+#include <cne_fib.h>              // for fib_create, fib_add...
+#include <cne_inet4.h>            // for inet_mtoh64 and inet_h64tom
+#include <cne_strings.h>          // for cne_strtok
+#include <net/cne_ether.h>        // for ether_addr
+
+#include "main.h"
+
+#define FIB_RULE_IP   0 /**< Index for IP address in a string array */
+#define FIB_RULE_MAC  1 /**< Index for MAC address in a string array  */
+#define FIB_RULE_PORT 2 /**< Index for TX port in a string array */
+
+struct ipv4_l3fwd_fib_rule {
+    uint32_t ip;
+    uint8_t depth;
+    struct ether_addr nh;
+    uint16_t tx_port;
+};
+
+static struct cne_fib *fib;
+
+static int
+l3fwd_fib_populate(struct fwd_info *fwd, struct cne_fib *fib)
+{
+    for (uint16_t i = 0; i < fwd->fib_size; i++) {
+        struct ether_addr *addr = NULL;
+        char *address[2];
+        struct ether_addr eaddr;
+        uint64_t eaddr_uint;
+        char *entry[3];
+        struct in_addr ip;
+        uint32_t ip_addr;
+        uint8_t mask;
+        uint64_t nexthop;
+        int tx_port;
+
+        /* Parse the comma seperated FIB entry */
+        if (cne_strtok(fwd->fib_rules[i], ",", entry, cne_countof(entry)) != 3)
+            CNE_ERR_RET("invalid number of fields for entry [orange]%u[]\n", i);
+
+        tx_port           = atoi(entry[FIB_RULE_PORT]);
+        jcfg_lport_t *dst = jcfg_lport_by_index(fwd->jinfo, tx_port);
+
+        if (!dst)
+            /* Cannot find a local port to match the entry */
+            CNE_ERR_RET("Invalid TX port index value [orange]%s[]\n", entry[FIB_RULE_PORT]);
+
+        /* Parse the IP address and mask */
+        if (cne_strtok(entry[FIB_RULE_IP], "/", address, cne_countof(address)) != 2)
+            CNE_ERR_RET("invalid number of ip and mask syntax\n");
+
+        if (inet_pton(AF_INET, address[0], (void *)&ip) != 1)
+            CNE_ERR_RET("Unable to convert IP4 address to network order\n");
+
+        ip_addr = ntohl(ip.s_addr);
+        mask    = (uint8_t)atoi(address[1]);
+
+        addr = cne_ether_aton(entry[FIB_RULE_MAC], &eaddr);
+        if (addr == NULL)
+            CNE_ERR_RET("Ethernet address is invalid [orange](%s)[]\n", entry[FIB_RULE_MAC]);
+        inet_mtoh64(&eaddr, &eaddr_uint);
+
+        /* Store both TX port and destination MAC in FIB's nexthop field. */
+        nexthop = ((uint64_t)tx_port << 48) | eaddr_uint;
+
+        if (cne_fib_add(fib, ip_addr, mask, nexthop) < 0)
+            CNE_ERR_RET(
+                "Failed to add FIB entry for IP address [orange]%s and lport [orange]%d[]\n",
+                address[0], tx_port);
+    }
+
+    return 0;
+}
+
+int
+l3fwd_fib_lookup(uint32_t *ip, struct ether_addr *eaddr, uint16_t *tx_port)
+{
+    uint64_t nexthop;
+
+    cne_fib_lookup_bulk(fib, ip, &nexthop, 1);
+    inet_h64tom(nexthop, eaddr);
+    *tx_port = (uint16_t)(nexthop >> 48);
+
+    return 0;
+}
+
+int
+l3fwd_fib_init(struct fwd_info *fwd)
+{
+    struct cne_fib_conf config;
+    config.max_routes = 1 << 16;
+    config.default_nh = 0xFFFFFFFFFFFF;
+    config.type       = CNE_FIB_DUMMY;
+
+    fib = cne_fib_create("l3fwd_fib", &config);
+    if (!fib)
+        CNE_ERR_RET("Failed to create FIB");
+
+    if (l3fwd_fib_populate(fwd, fib) < 0)
+        CNE_ERR_RET("Failed to populate FIB");
+    ;
+
+    /* Deallocate the entries now that we have the table populated. */
+    for (int i = 0; i < fwd->fib_size; ++i)
+        free(fwd->fib_rules[i]);
+    free(fwd->fib_rules);
+
+    return 0;
+}
--- a/examples/cndpfwd/main.c
+++ b/examples/cndpfwd/main.c
@@ -109,7 +109,7 @@ _fwd_test(jcfg_lport_t *lport, struct fwd_info *fwd)
     struct fwd_port *pd                          = lport->priv_;
     struct create_txbuff_thd_priv_t *thd_private = pd->thd->priv_;
     txbuff_t **txbuff;
-    int i, n_pkts;
+    int n_pkts;
 
     if (!pd)
         CNE_ERR_RET("fwd_port passed in lport private data is NULL\n");
@@ -120,7 +120,7 @@ _fwd_test(jcfg_lport_t *lport, struct fwd_info *fwd)
     if (n_pkts == PKTDEV_ADMIN_STATE_DOWN)
         return -1;
 
-    for (i = 0; i < n_pkts; i++) {
+    for (int i = 0; i < n_pkts; i++) {
         uint8_t dst_lport = get_dst_lport(pktmbuf_mtod(pd->rx_mbufs[i], void *));
         jcfg_lport_t *dst = jcfg_lport_by_index(fwd->jinfo, dst_lport);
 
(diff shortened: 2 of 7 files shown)


examples/cndpfwd/l3-fwd.c
```diff
@@ -0,0 +1,80 @@
...
+static struct cne_fib *fib;
+
+static int
+l3fwd_fib_populate(struct cne_fib *fib)
+{
+    /*
+     * 198.18.0.0/16 are set aside for RFC2544 benchmarking (RFC5735).
+     * Each FIB rule is IPv4, depth, next hop destination MAC, TX port.
+     */
+    struct ipv4_l3fwd_fib_rule ipv4_l3fwd_fib_rule_array[] = {
```
KeithWiles, line 30: Not a big fan of having these constant values in the source code. One option is to add a new key value pair array to the .jsonc file in the **options** section. This was done in the examples/cnet-graph jsonc file for adding socket connections.

I was thinking we can add a new array in the options section for this rules.

// fib-rules define the IP address, MAC address and port index for the FIB table entries
//    Three parts to each entry in the array. IP address with mask value/depth, MAC address and port index.
"fib-rules": [
    "198.18.0.0/24,02:00:01:02:03:04,1",
    "198.18.1.0/24,06:00:01:02:03:04,0",
]

The above is just an example and we can format the strings as you see fit. The example/cnet-graph has an example of how to access the options:graph:0-chanl strings.

Then we can remove hardcoded values from this file.

KeithWiles:
Excellent work, thanks.

KeithWiles:
It is a good idea to have the Git hook installed to clang-format the code as it is checked into the repo. This is what causes the check failures above.

The file cndp/.githooks/pre-commit needs to be copied to the correct location. Look in the CONTRIBUTING.md file around line 22 for more details.

Thank you for the patch it is looking good.

examples/cndpfwd/l3-fwd.c
stvml: We did this because it was a little easier to implement (and after discussion with @jeffreybshaw who pointed me to similar behavior in https://github.com/DPDK/dpdk/blob/main/examples/l3fwd/main.c#L204).

Totally agree that parsing the rules from config would be cleaner, though. Want us to do that on this PR, or open an enhancement issue once this one is merged?

stvml:
> It is a good idea to have the Git hook installed to clang-format the code as it is checked into the repo. This is what causes the check failures above.

Yep, read `CONTRIBUTING.md` but got excited and forgot to install hooks/actions when setting up our fork. :) Expect a `clang-format`ted update to this PR, probably sometime on Monday. Cheers!

examples/cndpfwd/l3-fwd.c
```diff
@@ -0,0 +1,80 @@
...
+
+    return 0;
+}
+
+int
+l3fwd_fib_lookup(uint32_t *ip, struct ether_addr *eaddr, uint16_t *tx_port)
+{
+    uint64_t nexthop;
+
+    cne_fib_lookup_bulk(fib, ip, &nexthop, 1);
```
stvml, line 57: While I'm here, also wanted to ask a maintainer's opinion on this call. `cne_fib_lookup_bulk` appears to be designed to handle arrays of `N` lookups, where in this case `N` is always 1. We call this function in the `for` loop over the array of received packets.

We didn't have time to look into this, but would it be more idiomatic if we were to set up those arrays in that RX loop, make a single call to `cne_fib_lookup_bulk`, then rewrite and TX the packets in a loop, as opposed to the current approach where we RX->lookup->rewrite->TX in a loop?

examples/cndpfwd/main.c
```diff
@@ -147,6 +147,56 @@ _fwd_test(jcfg_lport_t *lport, struct fwd_info *fwd)
...
+        rx_ip_hdr->hdr_checksum = htons(ntohs(rx_ip_hdr->hdr_checksum) + 1);
+        uint32_t ip_addr = ntohl(rx_ip_hdr->dst_addr);
+
+        l3fwd_fib_lookup(&ip_addr, &eaddr, &tx_port);
+
+        jcfg_lport_t *dst = jcfg_lport_by_index(fwd->jinfo, tx_port);
+
+        MAC_REWRITE(pktmbuf_mtod(pd->rx_mbufs[i], void *), &eaddr);
+
+        (void)txbuff_add(txbuff[dst->lpid], pd->rx_mbufs[i]);
```
stvml, line 182: Just noticed that we lost our logic here that should check whether the `jcfg_lport_by_index` call succeeded, which means this line could try to dereference NULL if the user specifies a nonexistent TX port index in their FIB definition.

The regular `fwd` test echoes packets back out of the RX port if there's no TX port matching the low byte of the MAC, but in our case, since the TX port indices are defined ahead of time, I'm inclined to drop them...

examples/cndpfwd/l3-fwd.c
KeithWiles: Have a look and see if you think opening a new PR or enhancement is reasonable, I would prefer it fixed in this PR. I can understand if you want to leave it this way and create a new PR or enhancement. If you are going to create a new PR please create an enhancement for the change.

examples/cndpfwd/main.c
KeithWiles: If we drop the packets with invalid TX port then we can not tell where the packet is dropped. Sending the packet back out the RX port seemed like a reasonable solution as we can see the packets getting looped back to the RX port. I can see both ways could work, let's drop them for now. Maybe some type of indicator to tell the user what is happening as it is an error.

When we add the jsonc **fib-rules**, we need to check the TX port exists and if not error out the rule.

The check for **dst** being null should be done, with a message could work. It could a be a huge number of messages, but I think it is reasonable.

examples/cndpfwd/l3-fwd.c
KeithWiles: I believe it would be a good change. It just means the arrays for the FIB lookup would need to be created dynamically using a dynamic array based on n_pkts or maximum statically created arrays. Using calloc or malloc here would be a performance hit.

stvml: Will create an enhancement issue for making this change once this PR is merged.

examples/cndpfwd/l3-fwd.c
```diff
@@ -0,0 +1,117 @@
...
+        for (uint16_t j = 1; j < 6; ++j)
+            mac_addr[j] = strtok(NULL, ":");
+
+        tx_port = atoi(entry[FIB_RULES_PORT]);
+
+        jcfg_lport_t *dst = jcfg_lport_by_index(fwd->jinfo, tx_port);
+
+        if (!dst)
+            /* Cannot find a local port to match the entry */
+            return -1;
```
KeithWiles, line 62: We can use the macro CNE_ERR_RET("Invalid TX port index value: %s\n", entry[FIB_RULES_PORT]); in place of return -1;

examples/cndpfwd/l3-fwd.c
```diff
@@ -0,0 +1,117 @@
...
+
+        uint32_t ip   = CNE_IPV4(atoi(ip_prefix[0]), atoi(ip_prefix[1]), atoi(ip_prefix[2]),
+                                 atoi(ip_prefix[3]));
+        uint8_t depth = (uint8_t)atoi(address[1]);
+
+        for (int j = 0; j < ETHER_ADDR_LEN; ++j) {
```
KeithWiles, line 68: We also have a MAC address convert routines?

```
struct ether_addr eaddr;
struct ether_addr *addr = cne_ether_aton(mac_addr, &eaddr);
if (addr == NULL)
    CNE_ERR_RET("Ethernet address is invalid (%s)\n", mac_addr);
```

examples/cndpfwd/l3-fwd.c
```diff
@@ -0,0 +1,117 @@
...
+            mac_le.ether_addr_octet[j] = (uint8_t)strtoul(mac_addr[j], NULL, 16);
+        }
+
+        inet_mtoh64(&mac_le, &eaddr);
```
KeithWiles, line 72: With the above this line is not required.

examples/cndpfwd/l3-fwd.c
```diff
@@ -0,0 +1,117 @@
...
+ * Copyright (c) 2022 Intel Corporation.
+ */
+
+#include <cne_fib.h>              // For fib_create, fib_add...
+#include <cne_inet4.h>            // for inet_mtoh64 and inet_h64tom
+#include <net/cne_ether.h>        // for ether_addr
+
+#include "main.h"
+
+#define FIB_RULES_IP   0 /**< JSON index for IP address in a single entry */
```
KeithWiles, line 11: Not really a JSON index. Maybe /**< Index to IP address in parsed string array */, ... or something like it.

examples/cndpfwd/l3-fwd.c
```diff
@@ -0,0 +1,117 @@
...
+            mac_le.ether_addr_octet[j] = (uint8_t)strtoul(mac_addr[j], NULL, 16);
+        }
+
+        inet_mtoh64(&mac_le, &eaddr);
+
+        /* Store both TX port and destination MAC in FIB's nexthop field. */
+        uint64_t nexthop = ((uint64_t)tx_port << 48) | eaddr;
+
+        if (cne_fib_add(fib, ip, depth, nexthop) < 0)
+            return -1;
```
KeithWiles, line 78: Use CNE_ERR_RET() instead of return -1;

examples/cndpfwd/main.c
```diff
@@ -147,6 +147,62 @@ _fwd_test(jcfg_lport_t *lport, struct fwd_info *fwd)
...
+    if (!pd)
+        CNE_ERR_RET("fwd_port passed in lport private data is NULL\n");
+
+    txbuff = thd_private->txbuffs;
+
+    n_pkts = __rx_burst(fwd->pkt_api, pd, pd->rx_mbufs, fwd->burst);
+    if (n_pkts == PKTDEV_ADMIN_STATE_DOWN)
+        return -1;
+
+    for (i = 0; i < n_pkts; i++) {
```
KeithWiles, line 169: Minor nit. On line 192 we use 'int i = 0', but here we use i = 0 with a global int i shadowing i. I would suggest to remove the global int i and change this one to int i = 0 to be consistent.

examples/cndpfwd/parse-args.c
```diff
@@ -76,6 +76,15 @@ process_callback(jcfg_info_t *j __cne_unused, void *_obj, void *arg, int idx)
                 if (f->xdp_uds == NULL)
                     CNE_ERR_RET("UDS handshake failed\n");
             }
+        } else if (!strncmp(obj.opt->name, FIB_RULES_TAG, nlen)) {
+            if (obj.opt->val.type == ARRAY_OPT_TYPE) {
+                f->fib_rules = malloc(obj.opt->val.array_sz * sizeof(char *));
```
KeithWiles, line 81: We normally use calloc() to make sure the memory is zero. May not matter much here, but just to be safe.

examples/cndpfwd/parse-args.c
```diff
@@ -76,6 +76,15 @@ process_callback(jcfg_info_t *j __cne_unused, void *_obj, void *arg, int idx)
...
+                uint16_t i   = 0;
```
KeithWiles, line 82: Move this line to after line 80 and remove the = 0; as it is not required. The forloop will set the value to zero.

Remove the {} from around the for loop as they are not required here and match the style.

examples/cndpfwd/l3-fwd.c
```diff
@@ -0,0 +1,117 @@
...
+        for (uint16_t j = 1; j < 3; ++j)
+            entry[j] = strtok(NULL, ",");
+
+        /* Parse the IP address and depth */
+        address[0] = strtok(entry[FIB_RULES_IP], "/");
+        address[1] = strtok(NULL, "/");
+
+        /* Parse the IP address */
+        ip_prefix[0] = strtok(address[0], ".");
+        for (uint16_t j = 1; j < 4; ++j) {
```
KeithWiles, line 47: We do not use {} around single like for, while, if, else, ... lines. Please remove if this is needed.

KeithWiles (changes requested):
Thank you for adding the JSON config support, but I did have a few minor changes and suggestions for error handling.

KeithWiles:
in the Smoke test above, we have a compile problem. Most likely a different compiler version.

```
../examples/cndpfwd/l3-fwd.c: In function ‘l3fwd_fib_init’:
../examples/cndpfwd/l3-fwd.c:113:17: error: invalid use of undefined type ‘struct cne_fib’
  113 |         free(fib->fib_rules[i]);
      |                 ^~
../examples/cndpfwd/l3-fwd.c:114:13: error: invalid use of undefined type ‘struct cne_fib’
  114 |     free(fib->fib_rules);
      |             ^~
```

examples/cndpfwd/parse-args.c
```diff
@@ -76,6 +76,16 @@ process_callback(jcfg_info_t *j __cne_unused, void *_obj, void *arg, int idx)
                 if (f->xdp_uds == NULL)
                     CNE_ERR_RET("UDS handshake failed\n");
             }
+        } else if (!strncmp(obj.opt->name, FIB_RULES_TAG, nlen)) {
+            uint16_t i;
+            if (obj.opt->val.type == ARRAY_OPT_TYPE) {
+                f->fib_rules = calloc(obj.opt->val.array_sz, sizeof(char *));
```
KeithWiles, line 82: Need to check to make sure f->fib_rules is not NULL from calloc().

KeithWiles (changes requested):
Thank you, one more minor issue.

KeithWiles (approved):
LGTM thank you

maryamtahhan:
@stvml @KeithWiles 
Hey guys
I'm seeing some pretty weird behaviour/performance with the scenario (cndpfwd in l3fwd mode between a physical port and veth):

                      +-------------cndpfwd-l3fwd---------+
traffic gen --> lport 1 (physical NIC) --> lport 2 (veth) ---------------+
                                                                                                                      | loopback
                                     traffic gen <-- lport 1 (physical NIC)  < --------+

Did ye notice any unusual performance drops in the phy to phy testing (I don't have enough ports to replicate phy to phy atm)?

stvml:
Honestly, the extent of our testing was veth to veth, and inside a VM... and we didn't have time to bring up a TRex instance and see what happens under realistic amounts of traffic. We just sent out single handmade Scapy packets to confirm that the FIB lookup and rewrite logic was doing what it was supposed to. 

Sorry we don't have anything useful to say on performance, but I'm thrilled that someone's tried putting this under load--what behavior are you seeing?

PR #255, opened 2023-01-04
update to clang-format 15.0.2

KeithWiles:
Update to clang-format version to 15.0.2 to fix a formatting issue.
Turn off clang-format in super-linter as it appears it is not the
correct version.

### DIFF
--- a/.github/workflows/super-linter.yml
+++ b/.github/workflows/super-linter.yml
@@ -31,7 +31,6 @@ jobs:
           # that .golangci.yml is available to the golangci-lint linter
           # https://github.com/github/super-linter/blob/main/docs/using-rules-files.md
           LINTER_RULES_PATH: /
-          VALIDATE_CLANG_FORMAT: true
           VALIDATE_BASH: true
           VALIDATE_GITHUB_ACTIONS: true
           VALIDATE_GO: true
--- a/.github/workflows/clang-format.yml
+++ b/.github/workflows/clang-format.yml
@@ -10,9 +10,9 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
-    - uses: DoozyX/clang-format-lint-action@v0.13
+    - uses: DoozyX/clang-format-lint-action@v0.15
       with:
         source: '.'
         exclude: './examples/vpp-plugin'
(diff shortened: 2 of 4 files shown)


.github/workflows/super-linter.yml
```diff
@@ -31,7 +31,7 @@ jobs:
           # that .golangci.yml is available to the golangci-lint linter
           # https://github.com/github/super-linter/blob/main/docs/using-rules-files.md
           LINTER_RULES_PATH: /
-          VALIDATE_CLANG_FORMAT: true
+          VALIDATE_CLANG_FORMAT: flase
```
jeffreybshaw, line 34: "false" is spelled incorrectly. But also, why not validate the clang format?

KeithWiles: I pushed a fix for the spelling error. The reason is appears the super-linter does not use the latest clang-format version or I could not find a way to have it use 15.0.2. The clang-format action does allow us to pick the correct version. I was trying to fix the problem in PR #250 as it works with clang-format version 15 on my machine.

I we can find a better solution and still use super-linter it would be great.

.github/workflows/super-linter.yml
```diff
@@ -31,7 +31,7 @@ jobs:
           # that .golangci.yml is available to the golangci-lint linter
           # https://github.com/github/super-linter/blob/main/docs/using-rules-files.md
           LINTER_RULES_PATH: /
-          VALIDATE_CLANG_FORMAT: true
+          #VALIDATE_CLANG_FORMAT: false # remove clang-format
```
jeffreybshaw, line 34: This is a weird way to do it. Why not just set it to "false"? Also I think we can get rid of the comment, it's pretty clear that you're not validating with clang format if you set it to false.

KeithWiles: The super-linter has a strange way of setting these values and you can not have mixture of true and false values.
https://github.com/marketplace/actions/super-linter#environment-variables
I can remove the comment.

jeffreybshaw (approved):

.github/workflows/super-linter.yml
jeffreybshaw: Weird!

PR #259, opened 2023-01-16
channel functions: update to check sysfs

maryamtahhan:
Update the functions that check the num of queues supported by a netdev to also check sysfs (e.g. veth). Also suppress the failure warning for retrieving the ring parameters when the operation is not supported.

Signed-off-by: Maryam Tahhan <mtahhan@redhat.com>

### DIFF
--- a/lib/core/osal/netdev_funcs.c
+++ b/lib/core/osal/netdev_funcs.c
@@ -249,7 +249,7 @@ netdev_get_channels_from_sysfs(const char *if_name, uint32_t *rxq, uint32_t *txq
     *rxq = *txq = 0;
 
     ret = snprintf(buf, PATH_MAX, "/sys/class/net/%s/queues/", if_name);
-    if (ret)
+    if (ret < 0)
         return;
 
     dir = opendir(buf);
@@ -272,7 +272,7 @@ netdev_get_channels(const char *ifname)
 {
     struct ifreq ifr;
     int fd, ret;
-    uint32_t ch;
+    uint32_t ch = 0;
     struct ethtool_channels eth_channels;
 
     fd = socket(AF_INET, SOCK_DGRAM, 0);
@@ -294,7 +294,7 @@ netdev_get_channels(const char *ifname)
          * try to get rx tx from sysfs, otherwise all traffic
          * is sent to a single stream, so max queues = 1.
          */
-        uint32_t rx_count, tx_count;
+        uint32_t rx_count = 0, tx_count = 0;
         netdev_get_channels_from_sysfs(ifr.ifr_name, &rx_count, &tx_count);
         ch = CNE_MAX(CNE_MAX(rx_count, tx_count), (uint32_t)1);
     } else {
--- a/lib/core/osal/netdev_funcs.c
+++ b/lib/core/osal/netdev_funcs.c
@@ -17,6 +17,8 @@
 #include <net/ethernet.h>
 #include <linux/ethtool.h>        // for ethtool_link_settings, ethtool_cmd, ETHT...
 #include <linux/sockios.h>        // for SIOCETHTOOL, SIOCGIFFLAGS, SIOCGIFHWADDR
+#include <dirent.h>
+#include <stdio.h>
 
 #include "netdev_funcs.h"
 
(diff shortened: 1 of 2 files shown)


lib/core/osal/netdev_funcs.c
```diff
@@ -236,11 +238,41 @@ netdev_get_offloads(const char *ifname, struct offloads *off)
...
+netdev_get_channels_from_sysfs(const char *if_name, uint32_t *rxq, uint32_t *txq)
+{
+    char buf[PATH_MAX];
+    struct dirent *entry;
+    DIR *dir;
+    int ret;
+
+    *rxq = *txq = 0;
+
+    ret = snprintf(buf, PATH_MAX, "/sys/class/net/%s/queues/", if_name);
```
KeithWiles, line 251: I think snprintf() returns the number of bytes in the string, which means the 'if' below would return without doing anything. I think the test should be if(ret >= PATH_MAX) CNE_RET("invalid path length\n");

lib/core/osal/netdev_funcs.c
```diff
@@ -236,11 +238,41 @@ netdev_get_offloads(const char *ifname, struct offloads *off)
...
+    if (ret)
+        return;
+
+    dir = opendir(buf);
+    if (!dir)
+        return;
```
KeithWiles, line 257: Not sure this return needs to be converted to CNE_RET() macro, but maybe it will just print out a message we do not need to see.

lib/core/osal/netdev_funcs.c
```diff
@@ -252,11 +284,28 @@ netdev_get_channels(const char *ifname)
...
+    if (ret) {
+        if (errno != EOPNOTSUPP) {
+            ret = -errno;
+            goto err;
+        }
+        /* If the device says it has no channels,
+         * try to get rx tx from sysfs, otherwise all traffic
+         * is sent to a single stream, so max queues = 1.
+         */
+        uint32_t rx_count, tx_count;
```
KeithWiles, line 297: Should these variables be set to zero or some know value?

lib/core/xskdev/xskdev.c
```diff
@@ -587,6 +592,35 @@ umem_create(lport_cfg_t *cfg)
...
+{
+    char buf[PATH_MAX];
+    struct dirent *entry;
+    DIR *dir;
+    int ret;
+
+    *rxq = *txq = 0;
+
+    ret = snprintf(buf, PATH_MAX, "/sys/class/net/%s/queues/", if_name);
+    if (ret)
```
KeithWiles, line 606: Same concern as the one above.

lib/core/xskdev/xskdev.c
```diff
@@ -587,6 +592,35 @@ umem_create(lport_cfg_t *cfg)
     return NULL;
 }
 
+static void
+xskdev_get_channels_from_sysfs(const char *if_name, uint32_t *rxq, uint32_t *txq)
```
KeithWiles, line 596: This function looks like a duplicate to the one in the netdev_funcs.c can we use the netdev_funcs.c here?

lib/core/xskdev/xskdev.c
```diff
@@ -600,33 +634,36 @@ xskdev_get_channel(const char *if_name, int *max_queues, int *combined_queues)
...
 
-    if (channels.max_combined == 0 || errno == EOPNOTSUPP) {
-        /* If the device says it has no channels, then all traffic
+        /* If the device says it has no channels,
+         * try to get channel info from sysfs, otherwise all traffic
          * is sent to a single stream, so max queues = 1.
          */
-        if (max_queues)
-            *max_queues = 1;
+        uint32_t rx_count, tx_count;
```
KeithWiles, line 651: Initialize the two variable may be required here.

KeithWiles (changes requested):
Thanks Maryam!

I had a couple questions and feedback.

lib/core/osal/netdev_funcs.c
maryamtahhan: I'll update it

lib/core/osal/netdev_funcs.c
maryamtahhan: yeah I can initialize them

lib/core/xskdev/xskdev.c
maryamtahhan: I will double check

lib/core/osal/netdev_funcs.c
maryamtahhan: yes I will update

lib/core/osal/netdev_funcs.c
maryamtahhan: cne_log depends on osal - so I can move things around to be able to do this - but it will involve some rejig of some files

lib/core/xskdev/xskdev.c
maryamtahhan: I moved xskdev.c to using the netdev funcs

KeithWiles (approved):
Thank you, look good to me.

lib/core/osal/netdev_funcs.c
KeithWiles: leave it for now

maryamtahhan:
@KeithWiles I thought this was the change that suppressed the warnings from netdev_get_ring_params... do we know what kernel version is being used where this error is still seen? maybe the return type isn't -EOPNOTSUPP ?

KeithWiles:
hmmm, The code will not print EOPNOTSUPP, so why are they seeing this message. Thanks.

Issue #221, opened 2022-10-20
Sharable Mempool between processes


KeithWiles:
**Difficulty**: Hard  to Medium

Modify CNDP [mempool code](https://github.com/CloudNativeDataPlane/cndp/tree/main/lib/core/mempool) to allow multiple processes to share a single mempool across processes. CNDP is a single process model application type and we need to be able to share buffers between processes without needing to share the entire process memory space which is what DPDK does today.

Sharing memory between processes is normally done with mmap() API to create a sharable region of memory. When sharing memory between processes you can not assume the virtual address of the memory space is the same between processes. That means memory pointers are not allowed in the shared memory region. This means the pointer currently being used need to be convert to offsets into the memory region.

TODO: **Add more information**

PR #261, opened 2023-01-18
Add MVP code for shared mempool and its example program

leonmatt:
This PR will add basic shared mempool code to the mempool library and targets issue #221 .

The way that this works is we mmap a region of shared memory in multiple processes at the same virtual address in each process and then we store all of the required mempool data structures in the region.

To test the program, build the repo, copy the run_proc.sh script from examples/shared_mempool into the build directory's shared_mempool artifacts, and then run "sudo ./run_proc.sh". The example should print out the addresses of the data structs of the mempool.

Next steps are to decide if we want to convert the pointers to offsets for AF_XDP and if we want to build out lockless rings within the mempool data.


KeithWiles:
@leonmatt this PR is getting out of date. I was thinking we can close this one and open it later if we need too. What do you think?

leonmatt:
@KeithWiles I agree. Perhaps if I find spare cycles, we can continue working on this in the future. In case someone picks this up in the future, I think we decided that it might be worthwhile to try placing the shared memory pool in the kernel address space, marking the pages r/w and accessible in User-Mode, and then writing a module that creates a memory policy so that the memory pool is only mapped into processes that CNDP is configured to have access.

PR #263, opened 2023-01-28
add hyperscan forwarding support

KeithWiles:
Add a simple forwarding mode using Hyperscan regex expressions to cndpfwd application. The example can be expanded to be a DDOS or DPI or a load balancer using Hyperscan. The example needs Hyperscan installed to work.

The cndpfwd application in Hyperscan mode will drop or forward packets to a different port. The Hyperscan string format is ```<ID>:/<regex>/<flags>``` flags are the standard Hyperscan flags.

The ID number can be in hex or decimal, but hex 0x00000000 is easier to use as the value is encoded with the port ID value in bits 16-23 , which is a value 0-N N being lport ID. The bits 24-31 can be used for flags and the only flag today is 0x80, which means the packets that match this expression should be dropped. The lower 16 bits is a unique ID value for each expression and is not used by the application.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -21,6 +21,18 @@ sudo apt-get update && sudo apt-get install -y \
     libpcap-dev meson pkg-config
 ```
 
+### Hyperscan support
+
+The cndpfwd example application supports a simple hyperscan forwarding type mode. The Hyperscan
+forwarding mode uses expressions in Hyperscan style in the jsonc file in options.hs_patterns array.
+The example can be expanded to be a DDOS or DPI or a load balancer using Hyperscan. If wanting to
+try Hyperscan you must install Hyperscan. Ubuntu has a package which can be installed to enable the
+build process to detect and use Hyperscan.
+
+```bash
+sudo apt-get install -y libhyperscan-dev # In Ubuntu 22.10 the Hyperscan version is 5.4.0
+```
+
 #### Optional packages needed to build documentation
 
 ```bash
--- a/examples/cndpfwd/fwd.jsonc
+++ b/examples/cndpfwd/fwd.jsonc
@@ -122,7 +122,7 @@
     //   no-restapi - (O) Disable RestAPI support
     //   cli        - (O) Enable/Disable CLI supported
     //   mode       - (O) Mode type [drop | rx-only], tx-only, [lb | loopback], fwd, tx-only-rx,
-    //                    acl-strict, acl-permissive
+    //                    acl-strict, acl-permissive, [hyperscan | hs]
     //   uds_path   - (O) Path to unix domain socket to get xsk map fd
     "options": {
         "pkt_api": "xskdev",
@@ -130,9 +130,19 @@
         "no-restapi": false,
         "cli": true,
         "mode": "drop",
+        // FIB rules <IP Address>,<Destination MAC>,<destination port>
         "l3fwd-fib-rules": [
             "198.18.0.0/24,02:00:01:02:03:04,1",
             "198.18.1.0/24,06:00:01:02:03:04,0"
+        ],
+        // The lport ID value (bits 16-23) is destination port, lower 16bits is unique ID value.
+        // If bits 31 is set then matching packets are dropped. See Hyperscan
+        // documentation for regex and flags values. <ID>:/<expression>/<flags>
+        // Any packets not matching a pattern below are dropped.
+        "hs-patterns": [
+            "0x80000000:/foobar/iH",
+            "0x00010001:/foo/i",
+            "0x00000002:/bar/i"
         ]
     },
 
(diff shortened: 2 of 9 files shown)


INSTALL.md
```diff
@@ -21,6 +21,17 @@ sudo apt-get update && sudo apt-get install -y \
     libpcap-dev meson pkg-config
 ```
 
+### Hyperscan support
+
+The cndpfwd example application supports a simple hyperscan forwarding type mode. The Hyperscan
```
maryamtahhan, line 26: HS used with cndpfwd is pretty much supporting a pattern matching permit/deny list?

examples/cndpfwd/fwd.jsonc
```diff
@@ -122,17 +122,27 @@
...
         "mode": "drop",
+        // FIB rules <IP Address>,<Destination MAC>,<destination port>
         "l3fwd-fib-rules": [
             "198.18.0.0/24,02:00:01:02:03:04,1",
             "198.18.1.0/24,06:00:01:02:03:04,0"
+        ],
+        // The lport ID value (bits 16-23) is destination port, lower 16bits is unique ID value.
+        // If bits 31 is set then matching packets are dropped. See Hyperscan
+        // documentation for regex and flags values. <ID>:/<expression>/<flags>
+        // Any packets not matching a pattern below are dropped.
```
maryamtahhan, line 141: Is this just true for the first rule with ID `0000`, the rest don't have bit 31 set?

examples/cndpfwd/parse-args.c
```diff
@@ -77,16 +77,27 @@ process_callback(jcfg_info_t *j __cne_unused, void *_obj, void *arg, int idx)
...
-
-                f->fib_size = i;
+                    f->fib_size++;
+                }
+            }
+        } else if (!strncmp(obj.opt->name, HS_PATTERN_TAG, nlen)) {
+            if (obj.opt->val.type == ARRAY_OPT_TYPE) {
+                f->hs_patterns = calloc(obj.opt->val.array_sz, sizeof(char *));
+                if (!f->hs_patterns)
+                    CNE_ERR_RET("Unable to allocate hsfwd information array\n");
```
maryamtahhan, line 94: Silly question - but should we check if umem memory has been mmap_alloc'd or if pd below has been alloced (or any other malloc'd variables) and do some cleanup before we return in error from the OPTIONS config processing?

examples/cndpfwd/hs-fwd.c
```diff
@@ -0,0 +1,258 @@
...
+        return -1;
+    }
+
+    for (int i = 0; i < fwd->hs_pattern_count; i++) {
+        int n = cne_strtok(fwd->hs_patterns[i], "/", entries, cne_countof(entries));
+
+        if (n < 2 || n > 3)
+            CNE_ERR_RET("Hyperscan pattern is invalid\n");
+
+        if (entries[0] == NULL)
```
maryamtahhan, line 184: is there a mismatch here between the index in the check `0` and the index in the CNE_ERR_RET statement `i` Or is the intention to print the fwd->hs_patterns[i] in the CNE_ERR_RET statement

examples/cndpfwd/hs-fwd.c
```diff
@@ -0,0 +1,258 @@
...
+            CNE_ERR_RET("Hyperscan pattern is invalid '%s'\n", entries[i]);
+        errno = 0;
+        id    = strtoul(entries[0], NULL, 0);
+        if (errno)
+            CNE_ERR_RET("Hyperscan pattern is ID for pattern: %s\n", strerror(errno));
```
maryamtahhan, line 189:
```suggestion
            CNE_ERR_RET("Hyperscan ID is invalid for pattern: %s error %s\n", fwd->hs_patterns[i], strerror(errno));
```

INSTALL.md
KeithWiles: I was thinking HS can provide a simple DDOS or DPI or even a load balancer, but this is far from a complete application for one of these types.

examples/cndpfwd/fwd.jsonc
KeithWiles: Correct, the rest are forwarding the packets to specific ports if they match a pattern.

examples/cndpfwd/parse-args.c
KeithWiles: Yes, I agree we should be doing a better job of cleaning up. I can add some code to cleanup the malloc'd memory in the new code, but we need to do a much better job of freeing memory everyplace or just reply on the OS to cleanup.

maryamtahhan: yeah that would work

maryamtahhan (approved):
LGTM - I will add an issue re the memory cleanup

PR #265, opened 2023-01-30
containerization: Enable podman + buildah

maryamtahhan:
Enable podman and buildah for CNDP container image creation and running.

Signed-off-by: Maryam Tahhan <mtahhan@redhat.com>

PR #266, opened 2023-02-01
docker workflow: update target

maryamtahhan:
Signed-off-by: Maryam Tahhan mtahhan@redhat.com

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -17,4 +17,4 @@ jobs:
       uses: actions/checkout@v2
 
     - name: Build Docker Image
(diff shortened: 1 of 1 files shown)


maryamtahhan:
depends on https://github.com/CloudNativeDataPlane/cndp/pull/265

KeithWiles (approved):

maryamtahhan:
Let's do this in a staged manner... I will restore the old docker build target in the other pull request and when that passes we can update the target

maryamtahhan:
after we update the target we can push another pull request then that removes the old target from the CNDP work base...

PR #281, opened 2023-03-21
k8s: fixup integration with the device plugin

maryamtahhan:
Changes were made in the environment variables pushed through the downward API to the CNDP pod. 
This patch updates at least the name of the devices and supports kind based deployment for testing.

Note: future work should update the number of queues detection


KeithWiles (approved):
LGTM

PR #284, opened 2023-03-27
Revert "k8s: fixup integration with the device plugin"

maryamtahhan:
Reverts CloudNativeDataPlane/cndp#281


maryamtahhan:
will submit everything in a single PR with the map pinning... reverting for now to fixup some of the linting issues picked up by github

PR #293, opened 2023-04-26
Add an example application of packet distributor

### LINKED ISSUE #287: Example Go application of packet distributor
In #223, we need some more examples to show the usage of [CNDP Go Binding](https://github.com/CloudNativeDataPlane/cndp/tree/main/lang/go/bindings) layer.

Here we track a detail example of packet distributor, it performs the distribution of packets that are received on an rx queue to different cores to process, and send to an tx queue to transmit.

haoruan:
#### What type of PR is this?
/kind feature

#### What this PR does / why we need it:
Add a real world example to show how to distribute packets to multiple workers.

#### Which issue(s) this PR fixes:
Fixes #287 

Signed-off-by: Hao Ruan <hao.ruan@intel.com>
Signed-off-by: Xiaotian Chen <xiaotian.chen@intel.com>

### DIFF
--- a/examples/cndpfwd/parse-args.c
+++ b/examples/cndpfwd/parse-args.c
@@ -25,6 +25,9 @@
 
 #include "main.h"        // for fwd_info, fwd, app_options, get_app_mode
 
+#define foreach_thd_lport(_t, _lp) \
+    for (int _i = 0; _i < _t->lport_cnt && (_lp = _t->lports[_i]); _i++, _lp = _t->lports[_i])
+
 static int
 process_callback(jcfg_info_t *j __cne_unused, void *_obj, void *arg, int idx)
 {
@@ -147,7 +150,6 @@ process_callback(jcfg_info_t *j __cne_unused, void *_obj, void *arg, int idx)
             /* Initialize a pktmbuf_info_t structure for each region in the UMEM space */
             pi = pktmbuf_pool_create(ri->addr, ri->bufcnt, obj.umem->bufsz, cache_sz, NULL);
             if (!pi) {
-                mmap_free(obj.umem->mm);
                 CNE_ERR_RET("pktmbuf_pool_init() failed for region %d\n", i);
             }
             snprintf(name, sizeof(name), "%s-%d", obj.umem->name, i);
@@ -171,6 +173,10 @@ process_callback(jcfg_info_t *j __cne_unused, void *_obj, void *arg, int idx)
             pd = calloc(1, sizeof(struct fwd_port));
             if (!pd)
                 CNE_ERR_RET("Failed to allocate fwd_port structure\n");
+
+            // Init lport to -1, so in cleanup routine we can know if we need to close it.
+            pd->lport = -1;
+
             lport->priv_ = pd;
 
             if (lport->flags & LPORT_SKB_MODE)
@@ -203,7 +209,6 @@ process_callback(jcfg_info_t *j __cne_unused, void *_obj, void *arg, int idx)
 
             pcfg.addr = jcfg_lport_region(lport, &pcfg.bufcnt);
             if (!pcfg.addr) {
-                free(pd);
                 CNE_ERR_RET("lport %s region index %d >= %d or not configured correctly\n",
                             lport->name, lport->region_idx, umem->region_cnt);
             }
@@ -218,14 +223,12 @@ process_callback(jcfg_info_t *j __cne_unused, void *_obj, void *arg, int idx)
             case XSKDEV_PKT_API:
                 pd->xsk = xskdev_socket_create(&pcfg);
                 if (pd->xsk == NULL) {
-                    free(pd);
                     CNE_ERR_RET("xskdev_port_setup(%s) failed\n", lport->name);
                 }
                 break;
             case PKTDEV_PKT_API:
                 pd->lport = pktdev_port_setup(&pcfg);
                 if (pd->lport < 0) {
-                    free(pd);
                     CNE_ERR_RET("pktdev_port_setup(%s) failed\n", lport->name);
                 }
                 break;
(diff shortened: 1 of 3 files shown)


lang/go/bindings/examples/distributor/pkg/distributor/cmd.go
```diff
@@ -0,0 +1,115 @@
...
+	"time"
+
+	"github.com/CloudNativeDataPlane/cndp/lang/go/bindings/cne"
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+)
+
+const (
+	version    = "v0.0.1"
```
KeithWiles, line 19: The version string is major.minor.patch, this means it should be 0.1.0 instead.

lang/go/bindings/examples/distributor/pkg/distributor/cmd.go
```diff
@@ -0,0 +1,115 @@
...
+
+var (
+	config  string
+	verbose bool
+)
+
+func NewDistributorCommand(ctx context.Context) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:    "distributor -c config.jsonc",
+		Short:  "distributor is an example to distribute packets",
```
KeithWiles, line 31: Update the short description to something like this:
'An example to distribute packets via lockless rings between worker threads'

lang/go/bindings/examples/distributor/pkg/distributor/cmd.go
```diff
@@ -0,0 +1,115 @@
...
+		}
+	}()
+
+	tx, err := NewTX(ctx, handle, lb)
+	if err != nil {
+		logrus.Errorf("failed to create tx: %s", err)
+		return
+	}
+	tx.Start()
+	defer tx.Stop()
```
KeithWiles, line 108: Should this line be moved to before tx.Start() line?

lang/go/bindings/examples/distributor/pkg/distributor/cmd.go
```diff
@@ -0,0 +1,115 @@
...
+
+	rx := NewRX(ctx, handle, lb)
+	rx.Start()
+	defer rx.Stop()
```
KeithWiles, line 112: Move defer line to be above rx.Start()

lang/go/bindings/examples/distributor/pkg/distributor/rx.go
```diff
@@ -0,0 +1,79 @@
...
+	for {
+		for _, pid := range lportIds {
+			select {
+			case <-rx.ctx.Done():
+				rx.wg.Done()
+				return
+			default:
+				size := cne.RxBurst(pid, packets)
+				for i := 0; i < size; {
+					i += rx.lb.Handle(packets[i])
```
KeithWiles, line 61: Does the handler function only return a one, it seems odd to have the handler return the increment value for the loop. It seems like the handler should only return an error or nothing. What happens if the handler returns zero, it can be stuck in this loop. I have not looked at the handler routine yet, so returning zero may not be an option.

lang/go/bindings/examples/distributor/pkg/distributor/tx.go
```diff
@@ -0,0 +1,92 @@
...
+			case <-tx.ctx.Done():
+				tx.wg.Done()
+				return
+			default:
+				n := tx.ring.DequeueN(items, numPackets)
+
+				for _, item := range items[:n] {
+					packet := (*cne.Packet)(unsafe.Pointer(item))
+
+					cne.TxBurst(pid, []*cne.Packet{packet}, true)
```
KeithWiles, line 69: The TxBurst() routine returns N or zero if the packets can not be sent. If a zero is returned the link is down and should we continue looping forever here? I think we need to handle the zero return case, maybe a message if we are in this loop for too long, because the link is down.

lang/go/bindings/examples/distributor/pkg/distributor/worker.go
```diff
@@ -0,0 +1,74 @@
...
+
+import (
+	"context"
+	"unsafe"
+
+	"github.com/CloudNativeDataPlane/cndp/lang/go/bindings/cne"
+)
+
+const (
+	initialBufferSize = 1024
```
KeithWiles, line 15: Should be named initialRingSize, as buffer is different IMO

KeithWiles (changes requested):
Looking pretty good, but I do have some comments.

Maybe you can also create a doc.go file to document the example and the distributor package. I know it was not done for the fwd example, but it will help others to understand the package and example better.

lang/go/bindings/examples/distributor/README.md
```diff
@@ -0,0 +1,41 @@
+# Packet Distributor
+
+A example application to distribute packets to workers.
+
+## Explanation
+
+This example contains 1 rx thread, 1 tx thread, and 4 worker threads.
+
+The rx thread receives packets, and then passes one packet to a worker via locklessring each time.
```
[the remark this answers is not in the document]
haoruan: Updated

lang/go/bindings/examples/distributor/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+The worker thread dequeues packets from its locklessring and just swaps packets' MAC address, and enqueues processed packets to another locklessring shared by all workers.
```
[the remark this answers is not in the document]
haoruan: Updated

lang/go/bindings/examples/distributor/config.jsonc
```diff
@@ -0,0 +1,155 @@
...
+    //    shared_umem - (O) Set to true to use xsk_socket__create_shared() API, default false
+    //    description | desc - (O) Description of the umem space.
+    "umems": {
+        "umem0": {
+            "bufcnt":64,
+            "bufsz": 2,
+            "mtype": "2MB",
+            "regions": [
+                32,
+                32
```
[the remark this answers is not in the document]
haoruan: Updated

lang/go/bindings/examples/distributor/README.md
```diff
@@ -0,0 +1,41 @@
...
+- Configure an ethtool filter to steer packets to a specific queue.
+  ```bash
+  sudo ethtool -N <devname> flow-type udp4 dst-port <dport> action <qid>
+  sudo ip link set dev <devname> up
+  ```
+- Edit `config.jsonc`, make sure the `lports` section has the same netdev name and queue id for which the ethtool filter is configured. Make sure the `threads` section has the correct `lports` configured
+
+## Usage
+```bash
+distributor is an example to distribute packets
```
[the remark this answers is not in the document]
haoruan: This is set in `cobra.Command`, I changed the description to "An example to distribute packets via lockless rings between worker threads"

lang/go/bindings/examples/distributor/cmd/distributor.go
```diff
@@ -0,0 +1,47 @@
+/* SPDX-License-Identifier: BSD-3-Clause
+ * Copyright(c) 2022-2023 Intel Corporation
```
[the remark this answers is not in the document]
haoruan: Updated

lang/go/bindings/examples/distributor/pkg/distributor/cmd.go
haoruan: Updated

lang/go/bindings/examples/distributor/pkg/distributor/cmd.go
haoruan: Updated

lang/go/bindings/examples/distributor/pkg/distributor/cmd.go
haoruan: It's common to place defer after resource or routine allocated/opened/created/started

lang/go/bindings/examples/distributor/pkg/distributor/rx.go
haoruan: `Handle` returns how many packets (1 or 0 in this case) has been distribute to workers, just like `EnqueueN(packet, 1)`

lang/go/bindings/examples/distributor/pkg/distributor/tx.go
haoruan: Updated, and will dump an error message if the loop takes more than 1 sec.

KeithWiles (approved):
LGTM

PR #314, opened 2023-05-25
Remove privileged lport flag from rust API.

manojgop:


KeithWiles (approved):
I assume this is correct not knowing Rust very well :-(

Thank you for updating the code.

PR #325, opened 2023-07-20
jsonc-gen.sh: fixup for multiple ports in a pod

maryamtahhan:
The qid was empty in the jsonc for any additional ports (after the first port) if we tried to generate the jsonc for multiple ports. This patch is an attempt to fix this.

Also LIST_OF_QIDS is no longer being set by the Device Plugin - so this was removed.


KeithWiles (approved):
Thanks

PR #326, opened 2023-07-21
txgen: reduce stats thread update to bring down cpu usage

maryamtahhan:
The txgen_stats() function was causing a full core to run at 100%... the cne_pause() function wasn't really alleviating the issue. I've opted to usleep here for 100ms as this is the refresh rate we also use for the cli prompt.


KeithWiles (approved):
Looks good

PR #329, opened 2023-07-28
cnet: free at the punt node

maryamtahhan:
There's no need to let the mbufs flow to the drop node from punt to be freed. It's also a little misleading as punted packets are being accounted for as dropped packets (IIUC).


KeithWiles (approved):
Thanks that was a good find.

PR #332, opened 2023-07-31
Cnet ether punt node

maryamtahhan:
Create a new punt node for ethernet frame. In the  case that the xsk bpf program doesn't filter only UDP/TCP traffic to cnet, then punt arp requests to the host kernel. This commit creates a new punt node called `punt_l2_kernel`. ARP requests are punted to it from the ptype node. The ARP request is then passed to the host through a tap device called `punt_ether`.

Maybe this can be combined with the IP packet punt node. But I went down this path to avoid having to do too many packet rechecks

### DIFF
--- a/lib/cnet/punt/punt_ether_kernel.c
+++ b/lib/cnet/punt/punt_ether_kernel.c
@@ -1,0 +1,185 @@
+/* SPDX-License-Identifier: BSD-3-Clause
+ * Copyright (c) Red Hat Inc.
+ * Copyright (c) 2023 Intel Corporation
+ */
+
+#include <net/cne_ether.h>           // for ether_addr_copy, cne_ether_hdr, ether_ad...
+#include <cnet.h>                    // for cnet_add_instance, cnet, per_thread_cnet
+#include <cnet_stk.h>                // for proto_in_ifunc
+#include <cnet_drv.h>                // for drv_entry
+#include <cnet_route.h>              // for
+#include <cnet_arp.h>                // for arp_entry
+#include <netinet/in.h>              // for ntohs
+#include <netpacket/packet.h>        // for sockaddr_ll
+#include <net/if.h>
+#include <net/if_arp.h>
+#include <linux/if_tun.h>
+#include <linux/if_ether.h>
+#include <sys/socket.h>
+#include <sys/ioctl.h>
+#include <stddef.h>        // for NULL
+#include <sys/types.h>
+#include <fcntl.h>
+#include <bsd/string.h>
+#include <sys/uio.h>
+
+#include <cne_graph.h>               // for
+#include <cne_graph_worker.h>        // for
+#include <cne_log.h>                 // for CNE_LOG, CNE_LOG_DEBUG
+#include <mempool.h>                 // for mempool_t
+#include <pktdev.h>                  // for pktdev_rx_burst
+#include <xskdev.h>
+#include <pktmbuf.h>        // for pktmbuf_t, pktmbuf_data_len
+#include <pktmbuf_ptype.h>
+#include <cne_vec.h>        // for
+#include <cnet_eth.h>
+#include <hexdump.h>
+#include <cnet_netif.h>        // for
+#include <netinet/if_ether.h>
+#include <pmd_tap.h>
+#include <cnet_node_names.h>
+#include "punt_ether_kernel_priv.h"
+
+#define PREFETCH_CNT 6
+
+static __cne_always_inline void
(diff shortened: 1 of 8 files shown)


lib/cnet/punt/punt_ether_kernel_priv.h
```diff
@@ -0,0 +1,66 @@
+/* SPDX-License-Identifier: BSD-3-Clause
+ * Copyright (c) Red Hat Inc.
+ * Copyright (c) 2023 Intel Corporation
+ */
+#ifndef __INCLUDE_PUNT_ETHER_KERNEL_PRIV_H__
+#define __INCLUDE_PUNT_ETHER_KERNEL_PRIV_H__
+
+#include <cne_common.h>
+#include <tun_alloc.h>
```
maryamtahhan, line 9:
```suggestion

```

lib/cnet/punt/punt_ether_kernel.c
```diff
@@ -0,0 +1,185 @@
...
+    return nb_objs;
+}
+
+static int
+punt_ether_kernel_node_init(const struct cne_graph *graph __cne_unused, struct cne_node *node)
+{
+    punt_ether_kernel_node_ctx_t *ctx = (punt_ether_kernel_node_ctx_t *)node->ctx;
+
+    lport_cfg_t cfg = {0}; /**< CFG for tun/tap setup */
+    ctx->mmap       = mmap_alloc(DEFAULT_MBUF_COUNT, DEFAULT_MBUF_SIZE, MMAP_HUGEPAGE_4KB);
```
maryamtahhan, line 123: Can probably reduce the number of buffers and descriptors as this is the exception path. WDYT?

lib/cnet/punt/punt_ether_kernel.c
```diff
@@ -0,0 +1,185 @@
...
+    if (ctx->mmap == NULL)
+        cne_panic("Failed to mmap(%lu, %s) memory",
+                  (uint64_t)DEFAULT_MBUF_COUNT * (uint64_t)DEFAULT_MBUF_SIZE,
+                  mmap_name_by_type(MMAP_HUGEPAGE_4KB));
+
+    memset(&cfg, 0, sizeof(cfg));
+
+    strlcpy(cfg.name, TAP_NAME, sizeof(cfg.name));
```
maryamtahhan, line 131: Could make the tap device name unique per cndp instance....

KeithWiles: I think it would be OK to have different names. I feel like I tried that, but not sure it was in this area.

I think with these changes we have two taps one tap at L2 and another after L4, is that correct?
It would be good to be able to see stats and multiple taps if possible.

KeithWiles (commented):
Looks pretty good. With the IPv6 support being added does it change that code some? I notice in punt_kernel.c it have IPv6 support added, but I am not that fluent in IPv6.

lib/cnet/punt/punt_ether_kernel.c
maryamtahhan: I think at L4 we weren't opening a tap? was it not just a socket that was being used to send the IPv4 packets down to the Kernel?

I guess we can try to add individual stats for the tap. I was just relying on gstats today to tell us how many packets went through the node and then by proxy how many went through the tap.

KeithWiles: We are using a socket to send packets to the kernel. As the code does not use tap here we can ignore the stats for now.

Are you ready for this PR to be reviewed and merged?

maryamtahhan: > We are using a socket to send packets to the kernel. As the code does not use tap here we can ignore the stats for now.
> 
> Are you ready for this PR to be reviewed and merged?

So this "punt ether" node uses a raw tap. I tried the socket approach for L2 and had to create a tap device anyway because in order to send the L2 frame through the socket I needed an ifindex for an interface. Which is why here I opted to just use a raw tap device. We can certainly look at adding specific stats no probs. 

I will add a separate PR for the documentation around this node. 

The L3 punt node that already exists just uses a socket.

maryamtahhan:
> I am not able to rebase this one as it states it has conflicts, but I do not see a button to resolve the conflicts. Do you see this too?

I tried a manual rebase there - git was not very happy :( but it all looks good

maryamtahhan:
nope - clang not happy :(

KeithWiles (approved):

maryamtahhan:
:( I really don't know why git is complaining. I will close this PR and create a new one based off the latest master as it's really not happy.

PR #333, opened 2023-07-31
fix build error due to missing declaration

pbanicev:
../lib/usr/app/cli/cli.h: In function 'get_root':                                                                                                                                                                                              
../lib/core/log/cne_log.h:344:13: error: implicit declaration of function 'unlikely' [-Werror=implicit-function-declaration]                                                                                                                   
  344 |         if (unlikely(!(exp)))

### DIFF
--- a/lib/core/log/cne_log.h
+++ b/lib/core/log/cne_log.h
@@ -20,6 +20,7 @@
 #include <cne_stdio.h>
 
 #include "cne_build_config.h"        // for CNE_ENABLE_ASSERT
+#include "cne_branch_prediction.h"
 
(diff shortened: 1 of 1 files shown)


KeithWiles (approved):
I do not see this error, but I do not think it will hurt to add the include file.

KeithWiles:
Please add a signed off line to the commit message as I can not commit it without that line. Thanks.

pbanicev:
Error is possible to trigger here with: option('enable_asserts', type: 'boolean', value: true,

KeithWiles (approved):
Thank you

PR #336, opened 2023-08-17
arp-request: send arp on all interfaces

maryamtahhan:
Change up the arp-request node to create an ARP request and send it out all cnet interfaces.

### DIFF
--- a/lib/cnet/arp/arp_request.c
+++ b/lib/cnet/arp/arp_request.c
@@ -13,7 +13,10 @@
 #include <netinet/in.h>           // for ntohs
 #include <stddef.h>               // for NULL
 #include <sys/types.h>
+#include <netpacket/packet.h>        // for sockaddr_ll
 #include <sys/socket.h>
+// #include <linux/if_arp.h>
+#include <netinet/if_ether.h>
 
 #include <cne_graph.h>               // for
 #include <cne_graph_worker.h>        // for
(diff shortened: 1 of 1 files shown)


lib/cnet/arp/arp_request.c
```diff
@@ -13,7 +13,10 @@
 #include <netinet/in.h>           // for ntohs
 #include <stddef.h>               // for NULL
 #include <sys/types.h>
+#include <netpacket/packet.h>        // for sockaddr_ll
 #include <sys/socket.h>
+// #include <linux/if_arp.h>
```
maryamtahhan, line 18: remove dead code and update copyright...

lib/cnet/arp/arp_request.c
```diff
@@ -204,7 +222,7 @@ arp_request_node_init(const struct cne_graph *graph __cne_unused, struct cne_nod
 {
     arp_request_node_ctx_t *ctx = (arp_request_node_ctx_t *)node->ctx;
 
-    ctx->s = socket(AF_INET, SOCK_RAW, IPPROTO_RAW);
+    ctx->s = socket(AF_PACKET, SOCK_DGRAM, htons(ETH_P_ARP));
     if (ctx->s < 0)
         CNE_ERR_RET("Failed to open RAW socket\n");
```
KeithWiles, line 227: open DataGram socket

maryamtahhan:
Going to close this PR. The socket is using the Kernel to send out the ARP requests on all interfaces attached to CNET. I can see the ARP reply come in and get forwarded to the Kernel thanks to the new L2 punt node. But the Kernel doesn't update the ARP tables. I also think that this type of socket requires elevated privileges and this is something we want to avoid for CNDP.

PR #340, opened 2023-08-31
Support for building latest quicly version

pbanicev:
This fixes building issues with latest quicly version.
Need to be run tested with some typical setup, see #334

PR #343, opened 2023-09-27
Quicly hotfix  pcb - fix chksum verify

maryamtahhan:
Depends on https://github.com/CloudNativeDataPlane/cndp/pull/340

Fixes the PCB lookup failure in the UDP input node for cnet and the failing csum verifications ...
Note: it looks like the same issue exists in the tcp input node also... so I applied the same fix

Note: Only tested UDP path with quicly client application...

### DIFF
--- a/lib/cnet/tcp/tcp_input.c
+++ b/lib/cnet/tcp/tcp_input.c
@@ -47,7 +47,9 @@ tcp_input_lookup(struct cne_node *node, pktmbuf_t *m, struct pcb_hd *hd)
     struct pcb_key key = {0};
     struct pcb_entry *pcb;
     struct cnet_metadata *md;
+#if 0
     uint16_t csum;
+#endif
     int a_family, a_len;
     struct pcb_entry *pcb2;
 
@@ -90,14 +92,14 @@ tcp_input_lookup(struct cne_node *node, pktmbuf_t *m, struct pcb_hd *hd)
     pcb = cnet_pcb_lookup(hd, &key, BEST_MATCH);
     if (likely(pcb)) {
         int rc = TCP_INPUT_NEXT_PKT_DROP;
-
+#if 0        // Disable till cne_ipv4_udptcp_cksum_verify is fixed
         if (is_pcb_dom_inet6(pcb))
             csum = cne_ipv6_udptcp_cksum_verify(&tip->ip6, &tip->tcp);
         else
             csum = cne_ipv4_udptcp_cksum_verify(&tip->ip4, &tip->tcp);
         if (csum)
             return rc;
-
+#endif
         m->userptr = pcb;
         in_caddr_copy(&md->faddr, &key.faddr); /* Save the foreign address */
         in_caddr_copy(&md->laddr, &key.laddr); /* Save the local address */
--- a/lib/cnet/tcp/tcp_input.c
+++ b/lib/cnet/tcp/tcp_input.c
@@ -31,25 +31,24 @@
 #include "tcp_input_priv.h"
 
 /* The TCP/IP Pseudo header */
-typedef struct tcpip_s2 {
+typedef struct l3_s2 {
     union {
-        struct cne_ipv4_hdr ip4; /* IPv4 header */
-        struct cne_ipv6_hdr ip6; /* IPv6 header */
+        struct cne_ipv4_hdr *ip4; /* IPv4 header */
+        struct cne_ipv6_hdr *ip6; /* IPv6 header */
     };
-    struct cne_tcp_hdr tcp; /* TCP header */
-} __cne_packed tcpip_t2;
+} __cne_packed l3_t2;
 
 static inline uint16_t
 tcp_input_lookup(struct cne_node *node, pktmbuf_t *m, struct pcb_hd *hd)
 {
     struct cnet *cnet = this_cnet;
-    tcpip_t2 *tip;
+    struct cne_tcp_hdr *tcp; /* TCP header */
+    void *l3;
+    l3_t2 tip;
     struct pcb_key key = {0};
     struct pcb_entry *pcb;
     struct cnet_metadata *md;
-#if 0
-    uint16_t csum;
-#endif
+    int16_t csum;
     int a_family, a_len;
     struct pcb_entry *pcb2;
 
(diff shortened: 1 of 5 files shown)


lib/cnet/tcp/tcp_input.c
```diff
@@ -90,14 +92,14 @@ tcp_input_lookup(struct cne_node *node, pktmbuf_t *m, struct pcb_hd *hd)
     pcb = cnet_pcb_lookup(hd, &key, BEST_MATCH);
     if (likely(pcb)) {
         int rc = TCP_INPUT_NEXT_PKT_DROP;
-
+#if 0        // Disable till cne_ipv4_udptcp_cksum_verify is fixed
         if (is_pcb_dom_inet6(pcb))
             csum = cne_ipv6_udptcp_cksum_verify(&tip->ip6, &tip->tcp);
         else
             csum = cne_ipv4_udptcp_cksum_verify(&tip->ip4, &tip->tcp);
```
maryamtahhan, line 99: [cne_ipv4_udptcp_cksum_verify](https://github.com/CloudNativeDataPlane/cndp/blob/5f0a018298e518aeb24940d5237731c532bf0013/lib/include/net/cne_ip.h#L340C2-L349C2) invokes: __ipv4_udptcp_cksum() which expects the udp checksum to be 0

Overall this function doesn't look right to me. I would've assumed that it would compare the incoming UDP checksum with a calculated value... @KeithWiles what do you think?

```
static inline int
cne_ipv4_udptcp_cksum_verify(const struct cne_ipv4_hdr *ipv4_hdr, const void *l4_hdr)
{
    uint16_t cksum = __ipv4_udptcp_cksum(ipv4_hdr, l4_hdr);

    if (cksum != 0xffff)
        return -1;

    return 0;
}
```

lib/cnet/tcp/tcp_input.c
```diff
@@ -90,14 +92,14 @@ tcp_input_lookup(struct cne_node *node, pktmbuf_t *m, struct pcb_hd *hd)
...
         if (csum)
```
maryamtahhan, line 100: Also csum here needs to be signed, at the fns above return -1 on a fail...

lib/cnet/tcp/tcp_input.c
KeithWiles: I thought when verifying the checksum if you run the checksum with the valid checksum in the header the result would be zero meaning the verify passed. Is that not the case?

lib/cnet/tcp/tcp_input.c
KeithWiles: Yes I agree it should be signed, but maybe the function should be returning a 0 or 1 for pass/fail?

KeithWiles (approved):
The changes look fine to me and I only had a couple comments.

lib/cnet/tcp/tcp_input.c
maryamtahhan: > I thought when verifying the checksum if you run the checksum with the valid checksum in the header the result would be zero meaning the verify passed. Is that not the case?

It does not seem to be....

maryamtahhan: would need to do a local test to verify the valid checksum case and try to dig down into what the issue...

For now - incoming packets from the quicly client seem to be failing...

lib/cnet/udp/udp_input.c
```diff
@@ -86,20 +90,22 @@ udp_input_lookup(pktmbuf_t *m, struct pcb_hd *hd)
...
 
     /* Create a 4x PCB lookup routine */
     pcb = cnet_pcb_lookup(hd, &key, BEST_MATCH);
     if (likely(pcb)) {
+#if 0        // Disable till cne_ipv4_udptcp_cksum_verify is fixed
         if ((pcb->opt_flag & UDP_CHKSUM_FLAG) && uip->udp.dgram_cksum) {
             if (is_pcb_dom_inet6(pcb))
                 csum = cne_ipv6_udptcp_cksum_verify(&uip->ip6, &uip->udp);
             else
                 csum = cne_ipv4_udptcp_cksum_verify(&uip->ip4, &uip->udp);
```
maryamtahhan, line 104: So this checksum function call won't work because uip->udp is not a pointer into the mbuf - that is one issue with the checksum verify... I've fixed this up locally. will push something soon. 

Even with the fix though the checksum check is failing... 
Wireshark seems to be happy that the incoming packets from quicly have a good udp checksum 
Screenshot 2023-09-28 at 10 13 37

maryamtahhan: Ok, so I've figured it out... I fixed up the csum verification to be actually carried out on the mbuf itself... Will post a patch on top of this PR

lib/cnet/tcp/tcp_input.c
maryamtahhan: So the issue was related to the layer 3 pointers we were using rather than the direct mbuf...

lib/cnet/tcp/tcp_input.c
maryamtahhan: maybe I should just rename the `csum` variable to something else as it's not really a csum returned by those functions

maryamtahhan:
I've only tested UDP and not TCP - would someone be able to check the TCP path?

lib/cnet/udp/udp_input.c
```diff
@@ -86,20 +90,22 @@ udp_input_lookup(pktmbuf_t *m, struct pcb_hd *hd)
     else
         key.laddr.cin_addr.s_addr = uip->ip4.dst_addr;
 
-    md->faddr.cin_port = be16toh(uip->udp.src_port);
-    md->laddr.cin_port = be16toh(uip->udp.dst_port);
+    key.faddr.cin_port = uip->udp.src_port;
```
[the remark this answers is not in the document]
maryamtahhan: will set md in case it's used later by the app...

maryamtahhan:
Screenshot 2023-09-28 at 12 36 23

lib/cnet/tcp/tcp_input.c
KeithWiles: Yes, I agree it is not a checksum. We can do that change now or after this one gets merged up to you or I can create a PR for that just that fix.

KeithWiles:
The code for TCP looks correct, but I can check it after the merge if that seems reasonable?

lib/cnet/tcp/tcp_input.c
maryamtahhan: I can push a separate PR to rename it after this no prob

PR #349, opened 2023-11-10
rust build: fixup failed workflow

maryamtahhan:
@manojgop Can you please review


maryamtahhan:
NVM - while this builds for me locally in an ubuntu container there seems to be a mismatch and doesn't build here. 

@manojgop you might have a look at fixing up the failing workflow...

PR #350, opened 2023-11-17
Hotfix rust

maryamtahhan:


KeithWiles (approved):
Thanks Maryam, I noticed some installs will it replace some libs in the system?

maryamtahhan:
> Thanks Maryam, I noticed some installs will it replace some libs in the system?



> Thanks Maryam, I noticed some installs will it replace some libs in the system?

That was the hope, I try all the steps locally in a container and the build passes. I'm a little puzzled why the build fails in the github job

maryamtahhan:
Nah - I think I've exhausted the path of trying to reproduce the working build from my env in the runner env... We need to get @manojgop to have a look

Sorry

PR #358, opened 2024-04-15
Fix dependency check for libnl-route-3.0 in main meson.build file

jalalmostafa:
The meson code to check if `libnl-route-3` is available or not duplicates the code for `libnl-cli-3.0`.
Instead of `nl_route_dep.found()`, the build file checks for `libnl-cli-3` (again).


KeithWiles:
Thank you for sending this PR. Could you add a DCO to the PR? If you go to the PR page and find the DCO line it has a link to suggest how to fit the problem.

KeithWiles (approved):
Thank you.

PR #367, opened 2024-05-14
`pktdev_close_all` should not call `pktdev_close` on unused pktdevs

### LINKED ISSUE #366: `pktdev_close_all` is counter-intuitive
`pktdev_close_all` should close all lports by iterating over all available lports.
However, as it is right now, it iterates over all `CNE_MAX_ETHPORTS` and calls `pktdev_close(int)` which fails if `!dev->data || !dev->dev_ops` is true and returns `EINVAL` then consequently `pktdev_close_all` also fails.

Here is a common scenario where this might happen: CNET graph has a `punt_l2_kernel` node which initiates a TUN/TAP pktdev port. At the end of the program, `cnet_stop` is called and all graph nodes are finalized including the `punt_l2_kernel` node which closes the TUN/TAP port. Now `pktdev_close_all` is called but fails because it tries to close a closed TUN/TAP port.

Such scenarios can leave open ports and cause undefined behaviors. I think we should do one of the following:

1. Ignore the case where `!dev->data || !dev->dev_ops` is true
2. Find a way where `pktdev_close_all` only iterates over unclosed ports e.g. using a flag to know if a port in the array `pktdev_devices` is open or not and only call `pktdev_close` for open ports
3. Remove `pktdev_close_all` API call all the way, as it is counter-intuitive and might cause confusion

For now, a workaround is to call `pktdev_close` only on lports in jcfg e.g. using `jcfg_foreach_lport`

jalalmostafa:
Closes #366

### DIFF
--- a/lib/core/pktdev/pktdev.c
+++ b/lib/core/pktdev/pktdev.c
@@ -249,6 +249,8 @@ int
 pktdev_close_all(void)
 {
     PKTDEV_FOREACH (i) {
+        if (pktdev_devices[i].state == PKTDEV_UNUSED)
+            continue;
         if (pktdev_close(i) < 0)
             return -1;
     }


KeithWiles (approved):
Looks good thanks.

PR #370, opened 2024-06-11
`chnl_connect` should accept operations from Channel 0

jalalmostafa:
The current implementation of `chnl_connect` rejects operations from channel 0 and returns `EFAULT`.
I did not understand the reason for this. Assume the CNDP application is a usual TCP client, then the first channel
created to initialize a connection could be channel 0 which is rejected.

https://github.com/CloudNativeDataPlane/cndp/blob/9413087ac193cabdbd3db2f6df64986232acb4df/lib/cnet/chnl/cnet_chnl.c#L486

Is this a typo (the developer wanted to check the channel data structure `ch` for NULL values instead of channel descriptor `cd`)? If so, this PR fixes the typo and does necessary refactoring.


KeithWiles (approved):
Thanks for finding this problem.

PR #372, opened 2024-06-19
tcp-chnl: add TCP connection established callback for connect and accept

jalalmostafa:
CNDP follows an asynchronous pattern. We cannot call some functions before the connection is established e.g. `chnl_send`, so we need a notification that the TCP connection has been established.
The commit adds calls to the TCP connection callback for clients when they establish a connection i.e. the connection is ready to send and receive data.

To simplify things, `CHNL_TCP_ACCEPT_TYPE` is renamed to `CHNL_TCP_ESTABLISHED_TYPE`. The callback type is used for both the server's `accept` and the client's `connect`. In the case of `accept`, the listening channel descriptor is passed to the callback. In the case of `connect`, the client channel descriptor is passed.

### DIFF
--- a/doc/guides/sample_app_ug/cnet_graph.rst
+++ b/doc/guides/sample_app_ug/cnet_graph.rst
@@ -219,13 +219,13 @@ The channel callback types are shown below:
 .. code-block:: c
 
     /** Channel callback types */
-    typedef enum {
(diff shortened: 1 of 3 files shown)


maryamtahhan (approved):
Seems reasonable to me

pktgen (approved):
Thanks

KeithWiles (approved):
Thanks for the fix

PR #374, opened 2024-07-08
pre-commit: integrate with cndp

maryamtahhan:
Install pre-commit for CNDP + its configuration.
Remove old githook.


maryamtahhan:
Sorry for the delay in this, I was away on some summer hols - fixed up the last outstanding issue and it's ready now

lang/rs/wireguard/patch/0001-Integrate-CNDP-AF-XDP-with-Rust-Wireguard.patch
```diff
@@ -91,7 +91,7 @@ index 640d8cb..7cb769b 100644
 @@ -5,12 +5,22 @@ license = "MIT"
  name = "wireguard-rs"
  version = "0.1.4"
- 
+
 +[[bin]]
```
KeithWiles, line 95: Changing the patch file could be a problem here, need to verify or we should ignore this file.

lang/rs/wireguard/patch/0003-Update-latest-crypto-library-version.patch
```diff
@@ -32,15 +32,15 @@ index 2941071..51cd887 100644
  dependencies = [
   "gimli",
  ]
- 
+
  [[package]]
```
KeithWiles, line 36: Same problem for this patch file.

KeithWiles (approved):
Going to approve, but I have a couple comments we may need to address.

We had a lot of mis-spelled words, which I most likely added :-)

This is a lot of work, thank you very much!

lang/rs/wireguard/patch/0001-Integrate-CNDP-AF-XDP-with-Rust-Wireguard.patch
maryamtahhan: Yeah, let me reset these and add them to the ignored files... sorry

PR #378, opened 2024-07-19
Update super-linter.yml

maryamtahhan:
github/super-linter@v6 is not allowed by the project settings.  Reverting till we allow it.


KeithWiles (approved):
Thank you

PR #388, opened 2024-07-30
chnl: add chnl_port to get port from connected channel

jalalmostafa:
This PR adds a new `chnl_port` API function, allowing us to know which port this channel is bound to.
It is important to allocate mbufs when sending data e.g.

```c
pktmbuf_t* bufs[256];
int portid = chnl_port(cd);
pktdev_buf_alloc(portid, bufs, burst);
chnl_send(cd, bufs, burst);
```

This can be easily done as the PCB entry will be linked to a specific port (interface and queue) in the FIB when the connection has been established `chnl_accept` or `chnl_connect` has finished.


KeithWiles (approved):
Thank you

PR #390, opened 2024-08-05
cnet-graph: do not forward TCP control packets to chnl_callback

jalalmostafa:
CNET-Graph will send TCP control packets sometimes e.g. SYN, FIN, ACK, etc. We do not need to notify these packets. Instead, we only notify packets that have user data. For UDP, we notify all.

Let me know your comments how the decision to forward the packet to chnl_callback can be improved.
Right now I check data length. How about we check `PSH` or `URG` flags? If they are set in the tcb, then the packet has user data and we forward to `chnl_callback`. I am rethinking it because I am thinking of traffic generators that generate empty segments (where data len is zero), we may still need to get notifications in `chnl_callback` for these.

### DIFF
--- a/lib/cnet/eth/eth_tx.c
+++ b/lib/cnet/eth/eth_tx.c
@@ -17,8 +17,27 @@
 #include <cnet_node_names.h>
 #include "eth_tx_priv.h"        // for eth_tx_node_ctx_t, ETH_TX_NEXT_MAX
 #include "chnl_callback_priv.h"
+#include "cnet_pcb.h"
 
 static struct eth_tx_node_main eth_tx_main;
+
+static inline int
+eth_tx_node_pktmbuf_has_userdata(pktmbuf_t *mbuf)
+{
+    struct pcb_entry *pcb = mbuf->userptr;
+    if (unlikely(!pcb))
+        return -1;
+
+    if (pcb->ip_proto == IPPROTO_UDP)
+        return ETH_TX_NEXT_PKT_CALLBACK;
+
+    if (pcb->ip_proto == IPPROTO_TCP)
+        return mbuf->data_len > (mbuf->l2_len + mbuf->l3_len + mbuf->l4_len)
+                   ? ETH_TX_NEXT_PKT_CALLBACK
+                   : -1;
+
+    return -1;
+}
 
 static uint16_t
 eth_tx_node_process(struct cne_graph *graph, struct cne_node *node, void **objs, uint16_t nb_objs)
@@ -27,7 +46,15 @@ eth_tx_node_process(struct cne_graph *graph, struct cne_node *node, void **objs,
     uint16_t port          = ctx->port; /* Get TX port id */
     uint16_t count         = nb_objs;
 
-    CNE_SET_USED(graph);
+    pktmbuf_t *mbuf0, *mbuf1, *mbuf2, *mbuf3, **pkts;
+    int next0, next1, next2, next3;
+    void **to_next, **from;
+    uint16_t n_left_from;
+    uint16_t held = 0;
+
+    pkts        = (pktmbuf_t **)objs;
+    from        = objs;
+    n_left_from = nb_objs;
 
     if (nb_objs) {
         do {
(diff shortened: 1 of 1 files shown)


lib/cnet/eth/eth_tx.c
```diff
@@ -17,17 +17,44 @@
...
 static struct eth_tx_node_main eth_tx_main;
 
+static inline int
+eth_tx_node_pktmbuf_has_userdata(pktmbuf_t *mbuf)
+{
+    struct pcb_entry *pcb = mbuf->userptr;
+    if (unlikely(!pcb))
+        return -1;
+
+    if (pcb->ip_proto == IPPROTO_UDP)
```
KeithWiles, line 31: If we remove these two lines and at the bottom of the routine we return ETH_TX_NEXT_PKT_CALLBACK. This will reduce the number of 'if' tests and if a packet is not TCP with data the packet will be sent to the callback.

Does that make sense to only look at TCP packets and everything else can be sent to the callback?
Still thinking about this change as it is trying to optimize for a traffic generator use case.

KeithWiles:
I am concerned if we restrict some TCP packets from going to the callback the TCP protocol will not be handled correctly by the Kernel stack. It is valid to have zero data TCP packets with flags like when closing or other reasons.

The PR is also addressing a traffic generator or DOS attack use case.

lib/cnet/eth/eth_tx.c
jalalmostafa: What about other protocols e.g. ICMPv6? 
Do you think we should report control protocols to the user application?
My idea is we do not need to report control protocols because they should be handled by the stack itself, just like in the linux kernel: the user space only gets user data (regardless of zero size user packets like traffic generators)

jalalmostafa:
Hi @KeithWiles 
Thanks for the feedback. That's true but closing is handled elsewhere in the chnl_callback (`CHNL_UDP_CLOSE_TYPE` and `CHNL_TCP_CLOSE_TYPE`). I think all the other conditions should either be handled using dedicated callback types if they are needed or just be handled inside the CNET stack without user intervention just like in the Linux kernel. This was also the usual case before adding the `CHNL_*_SENT_TYPE` callback types.

KeithWiles:
Could you please run the 'graph dot` command and paste one of the text files here. I want to see the graph layout again.

Still thinking about this change, but so far I think it maybe OK.

jalalmostafa:
```dot
digraph cndp_tcpecho_graph_1 {
	rankdir=LD; bgcolor=mistyrose
	label=<<font color='black'>Graph Name: </font><font color='blue' point-size='20'> <b>cndp_tcpecho_graph_1</b></font>      <font color='black'><b>(pkt_drop removed)</b></font>>
	edge [color=blue, arrowsize=0.6]
	node [margin=0.1 fontcolor=black fontsize=16 width=0.8 shape=box color=black style="filled,rounded"]
	{"eth_rx-0" [fillcolor=lavender]}->"ptype"
	{"eth_tx-0" [fillcolor=cyan]}->"chnl_callback"
	{"eth_rx-1" [fillcolor=lavender]}->"ptype"
	{"eth_tx-1" [fillcolor=cyan]}->"chnl_callback"
	{"ip4_input" [fillcolor=mediumspringgreen]}->{"ip4_forward" "ip4_proto"}
	{"ip4_output" [fillcolor=mediumspringgreen]}->{"arp_request" "eth_tx-0" "eth_tx-1"}
	{"ip4_forward" [fillcolor=mediumspringgreen]}->{"arp_request" "eth_tx-0" "eth_tx-1"}
	{"ip4_proto" [fillcolor=mediumspringgreen]}->{"udp_input" "tcp_input"}
	{"udp_input" [fillcolor=cornsilk]}->{"chnl_recv" "punt_kernel"}
	{"udp_output" [fillcolor=cornsilk]}->"ip4_output"
	{"tcp_input" [fillcolor=lightpink]}->{"chnl_recv" "punt_kernel"}
	{"tcp_output" [fillcolor=lightpink]}->"ip4_output"
	{"chnl_callback" [fillcolor=lightgrey]}
	{"chnl_recv" [fillcolor=yellowgreen]}->"chnl_callback"
	{"ptype" [fillcolor=goldenrod]}->{"punt_kernel" "punt_l2_kernel" "ip4_input" "gtpu_input"}
	{"arp_request" [fillcolor=mediumspringgreen]}
	{"punt_kernel" [fillcolor=coral]}
	{"punt_l2_kernel" [fillcolor=coral]}
	{"gtpu_input" [fillcolor=lightskyblue]}->"ip4_input"
}
```
graph

KeithWiles:
Thanks for the picture.

I was thinking about the PSH and URG flag test, you can have PSH (I believe) without user data and seems like checking only the size is reasonable.

KeithWiles (approved):
Thanks for the PR

jalalmostafa:
Thanks @KeithWiles !

PR #420, opened 2025-04-08
fix: build on latest ubuntu

maryamtahhan:
Fixes error: 'calloc' sizes specified with 'sizeof' in the earlier argument and not in the later argument [-Werror=calloc-transposed-args] in docker build


KeithWiles (approved):
A sed guru :-) thanks

